Issue #1, opened 2018-02-14
groups with 4 rows and groups with 3 rows


Hagellach37:
When creating the groups for project 13505 (Swipe to help MSF in Douentza, Mali (1)) some groups (31 out of 1138) where not created correctly.

![grafik](https://user-images.githubusercontent.com/7045979/36232453-2048d55c-11e2-11e8-95be-1e9f3ba74442.png)

This file contains the geometries of the groups:
[groups_13505.zip](https://github.com/mapswipe/python-mapswipe-workers/files/1725912/groups_13505.zip)

Hagellach37:
removed incorrect groups to prevent users from contributing to these groups. There is still plenty of other groups. :)

PR #2, opened 2018-02-15
fix groups height issue #1

Hagellach37:
The grouping algorithm was adjusted.
We avoid calculating the Tile coordinates from the geometries all the time, and just do it initially. Afterwards, tile coordinates are derived from the coordinates of the previous group.


import_module/create_groups.py
```diff
@@ -254,25 +262,31 @@ def get_vertical_slice(geomcol, zoom):
         pixel = lat_long_zoom_to_pixel_coords(ymax, xmin, zoom)
         tile = pixel_coords_to_tile_address(pixel.x, pixel.y)
         TileX_left = tile.x
-        TileY_top = tile.y
+
+        # this is a fix for incorrect groups height
```
ericboucher: Can you be a bit more precise around what the fix is? What is happening?

Issue #11, opened 2018-03-12
'ZeroDivisionError' during project import


Hagellach37:
during the import a zero division error occured in the import workflow.

PR #12, opened 2018-03-12
fix zero division error, add check and also fix multipolygon issue #11

Hagellach37:
I looked at the groups creation and added few lines to avoid zero division errors.

### DIFF
--- a/import_module/create_groups.py
+++ b/import_module/create_groups.py
@@ -141,7 +141,12 @@ def get_geometry_from_file(infile):
 
 def get_horizontal_slice(extent, geomcol, zoom):
     logging.info('geomcol: %s' % geomcol)
-    slice_collection = ogr.Geometry(ogr.wkbGeometryCollection)
+
+    slice_infos = {
+        'tile_y_top': [],
+        'tile_y_bottom': [],
+        'slice_collection': ogr.Geometry(ogr.wkbGeometryCollection)
+    }
 
     xmin = extent[0]
     xmax = extent[1]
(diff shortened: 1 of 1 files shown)


import_module/create_groups.py
```diff
@@ -205,11 +209,15 @@ def get_horizontal_slice(extent, geomcol, zoom):
...
+                slice_infos['tile_y_bottom'].append(TileY+3)
+                slice_infos['slice_collection'].AddGeometry(sliced_poly)
                 #logging.info('sliced poly: %s' % sliced_poly)
             elif sliced_poly.GetGeometryName() == 'MULTIPOLYGON':
                 for geom_part in sliced_poly:
-                    slice_collection.AddGeometry(geom_part)
+                    slice_infos['tile_y_top'].append(TileY)
+                    slice_infos['tile_y_bottom'].append(TileY + 3)
+                    slice_infos['slice_collection'].AddGeometry(geom_part)
                     #logging.info('sliced poly: %s' % sliced_poly)
```
ericboucher: Would these be helpful to debug? If so, maybe switch to logging.debug and uncomment?

import_module/create_groups.py
```diff
@@ -261,38 +271,32 @@ def get_vertical_slice(geomcol, zoom):
...
-            TileY_bottom += 3
-
         # get lower right tile coordinates
         pixel = lat_long_zoom_to_pixel_coords(ymin, xmax, zoom)
         tile = pixel_coords_to_tile_address(pixel.x, pixel.y)
         TileX_right = tile.x
 
+
+        # we don't compute tile y top and tile y botton coordinates again, but get the ones from the list
+        # doing so we can avoid problems due to rounding errors and resulting wrong tile coordinates
```
ericboucher: and resulting IN wrong tile coordinates

import_module/create_groups.py
Hagellach37: good question. I think, I never used this info in the last times. so maybe I will just remove it.

import_module/create_groups.py
Hagellach37: +1

PR #14, opened 2018-03-14
Introduce Travis.ci and test docker

ericboucher:
Update module usage

Add Travis config


.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1 @@
+{}
```
mathcass: Is this something that we should ignore? I feel like this is a cache of previous test failures. The standard Python gitignore ignores this directory: https://github.com/github/gitignore/blob/master/Python.gitignore#L48

.travis.yml
```diff
@@ -0,0 +1,24 @@
+language: python
```
mathcass: Thanks for doing the legwork to get Travis running. Any time I've set it up in the past, it's taken a *lot* of trial an error.

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+sudo: required
+python:
+  - "3.6"
+# command to install dependencies
+# install gdal
+before_install:
+  - sudo apt-get update -y
+  - sudo apt-get install -y build-essential
+  - sudo apt-get install -y gcc
+  - sudo apt-get install -y gdal-bin
```
mathcass: Is there a way to log out the version of GDAL that this installs, for informational purposes. I don't feel the need to pin this based on what @Hagellach37 has said, but I do think it'd be helpful to look back through tests to see which version it is

export_module/run_export.py
```diff
@@ -1,32 +1,28 @@
...
 
 # define arguments that can be passed by the user
 parser = argparse.ArgumentParser(description='Process some integers.')
 parser.add_argument('-l', '--loop', dest='loop', action='store_true',
-                    help='if loop is set, the import will be repeated several times. You can specify the behaviour using --sleep_time and/or --max_iterations.')
+                    help='if loop is set, the import will be repeated several times.'
+                         'You can specify the behaviour using --sleep_time / --max_iterations.')
 parser.add_argument('-s', '--sleep_time', required=False, default=None, type=int,
-                    help='the time in seconds for which the script will pause in beetween two imports')
+                    help='ime in seconds for which the script will pause in beetween two imports.')
```
mathcass: I think this might be a typo since you changed 'the time' to 'ime'

setup.py
```diff
@@ -0,0 +1,11 @@
+from setuptools import setup, find_packages
```
mathcass: Thanks for turning this into a package, I think that really streamlines things. 

Could you add something to the README about how to install it? I think just the standard `python setup.py install` should be fine

setup.py
```diff
@@ -0,0 +1,11 @@
...
+
+setup(
```
mathcass: I think you might need an [`install_requires`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies) in here. Some folks [use the `requirements.txt`](https://stackoverflow.com/questions/14399534/reference-requirements-txt-for-the-install-requires-kwarg-in-setuptools-setup-py) that they included

setup.py
mathcass: Also, do you see a difference in the way deployment would happen because of this? I would think that deployment looks more like a `git pull && python setup.py install` rather than just re-pulling

Hagellach37:
Nice work. I must say, that I have no clue how to use Travis. But I'm happy if you could introduce me to the concept behind. :)
I worked with docker just the recent weeks quite frequently to set up the disaster openroute service.

ericboucher:
It's kind of a WIP, but I just wanted to have Docker to mimick the test
behavior on Travis. But I think it makes sense to dockerize the repo
afterwards, this way we don't need to rely on people installing / debugging
dependencies.

Travis is a simple CI tool and they are free for open source projects, so
quite convenient in our case!

2018-03-14 10:23 GMT-07:00 Hagellach37 <notifications@github.com>:

docker/docker-compose.yaml
```diff
@@ -5,6 +5,6 @@ services:
         context: .
     volumes:
       # Copy complete folder.
-      - ./:/home/mapswipe-workers/:rw
+      - .././:/home/mapswipe-workers/:rw
```
mathcass: I'm curious, why was `.././` needed rather than just `../`

mathcass:
Is this still worth doing?

PR #16, opened 2018-03-14
Fix transfer results scalability (fixes #15)

### LINKED ISSUE #15: transfer results script is too slow for many results
When many results are uploaded to the mapswipe database, we may not be able to import them quick enough.

Potential bottlenecks are:
* insert into mysql database and check for duplicates
* remove entries in firebase

Hagellach37:
This fixes an issue that occured during the last days.
Our old transfer results script was not able to import the results to mysql as quick as new data came into firebase. This was really critical and our server was using 100% all the time.

In the old script each result was transfered to mysql individually and afterwards deleted in firebase. This was not scalable.

In the new script the results from firebase are saved into a textfile. This textfile is first uploaded to an empty table in mysql ('raw_results'), afterwards we insert the items from the raw results table to the results table and check for duplicates.
Finally all entries are deleted in firebase using the "update" function. Removing items individually was not fast enough.

### DIFF
--- a/transfer_results_module/transfer_results.py
+++ b/transfer_results_module/transfer_results.py
@@ -73,6 +73,8 @@ def delete_firebase_results(all_results):
             data[task_id][child_id] = None
             q.put([fb_db, task_id, child_id])
 
+        print(data[task_id])
+
     for i in range(num_threads):
         worker = threading.Thread(
             target=delete_result_in_firebase,
--- a/transfer_results_module/transfer_results.py
+++ b/transfer_results_module/transfer_results.py
@@ -9,22 +9,16 @@ import sys
 sys.path.insert(0, '../cfg/')
 sys.path.insert(0, '../utils/')
 
+import error_handling
 import logging
 import json
-import numpy as np
 import os
-import threading
 import time
-from queue import Queue
-
 import requests
-
-import pymysql
 from auth import firebase_admin_auth
 from auth import mysqlDB
-from send_slack_message import send_slack_message
-
 import argparse
+
 parser = argparse.ArgumentParser(description='Process some integers.')
 parser.add_argument('-l', '--loop', dest='loop', action='store_true',
                     help='if loop is set, the import will be repeated several times. You can specify the behaviour using --sleep_time and/or --max_iterations.')
@@ -42,29 +36,12 @@ def get_results_from_firebase():
     results = fb_db.child("results").get().val()
     return results
 
-def delete_result_in_firebase(q):
-
-    while not q.empty():
-        fb_db, task_id, child_id = q.get()
-        try:
-            fb_db.child("results").child(task_id).child(child_id).remove()
-            print('deleted entry: %s, %s' % (task_id, child_id))
-        except:
-            # add a catch, if something with the connection to firebase goes wrong and log potential errors
-            tb = sys.exc_info()
-            logging.warning(str(tb))
-
-        q.task_done()
-
 
 def delete_firebase_results(all_results):
     firebase = firebase_admin_auth()
     fb_db = firebase.database()
 
-    # we will use a queue to limit the number of threads running in parallel
-    q = Queue(maxsize=0)
-    num_threads = 24
-
+    # we will use multilocation update to delete the entries, therefore we crate an dict with the items we want to delete
     data = {
         'results': {}
     }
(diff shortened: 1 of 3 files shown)


Hagellach37:
I'm testing the scripts currently on the server. Looks fine.

transfer_results_module/transfer_results.py
```diff
@@ -45,93 +38,114 @@ def get_results_from_firebase():
...
+
+    number_of_results = 0
     for task_id, results in all_results.items():
-        # there might be several results for a task_id
         for child_id, result in results.items():
-            # each result has a user_id and data
-            data = result['data']
-            q.put([task_id, child_id, data])
+            number_of_results += 1
+            outline = '{task_id}\t{user_id}\t{project_id}\t{timestamp}\t{result}\t{wkt}\t{task_x}\t{task_y}\t{task_z}\t{duplicates}\n'.format(
```
mathcass: It looks like you're wanting to write this out as tab-delimited content? I would tend to prefer using something like the [`csv`](https://docs.python.org/3.6/library/csv.html) module for stuff like this (you can specify TAB as the delimiter). In particular, the `DictWriter` is really convenient for stuff like this where you want to export all of the data like this. 

That said, this would probably be a little bit cleaner (and easier to maintain): 

```
outline = '\t'.join([
    task_id,
    result['data']['user'],
    result['data']['projectId'],
    result['data']['result'],
    result['data']['timestamp'],
    result['data']['wkt'],
    task_id.split('-')[1],
    task_id.split('-')[2],
    task_id.split('-')[0],
    0])
outline += '\n'

```

Note that this also requires that everything is a string. You could do this as well:

```
output_list = [
    task_id,
    result['data']['user'],
    result['data']['projectId'],
    result['data']['result'],
    result['data']['timestamp'],
    result['data']['wkt'],
    task_id.split('-')[1],
    task_id.split('-')[2],
    task_id.split('-')[0],
    0
]


outline = '\t'.join([str(s) for s in output_list])
outline += '\n'

```

transfer_results_module/transfer_results.py
```diff
@@ -146,19 +160,21 @@ def run_transfer_results():
...
     if os.path.isfile(results_filename):
-        logging.warning("there are results in %s that we didnt't insert. do it now!" % results_filename)
-        print("there are results in %s, that we didnt't insert. do it now!" % results_filename)
         # start to import the old results first
         with open(results_filename) as results_file:
             results = json.load(results_file)
-            transfer_results(results_filename, results)
+            results_txt_filename = results_to_txt(results)
+            logging.warning("there are results in %s that we didnt't insert. do it now!" % results_filename)
+            save_results_mysql(results_txt_filename)
```
mathcass: I think there might be some sneaky issues that could come up with this that are worth outlining and trying to account for. Namely, what happens if something goes wrong within `save_results_mysql`? An intermittent connection issue, timeout, or something else could cause this worker to be unable to save the results properly. Then, the workflow would be:

* Put results in text file
* Try to save results to MySQL
  * Saving to MySQL fails
* Delete the results in the file

which means we just lost everything. 

To make it more resilient, it'd be better to have `save_results_mysql` raise some sort of exception so that you can catch it. If things go "as planned" then proceed to delete the results from the file. Otherwise, you'll want to bail and not delete the result file until you can properly save them.

transfer_results_module/transfer_results.py
ericboucher: I agree that using the csv lib is cleaner and will make catching error easier. This way, you can be more certain that your output file will actually be usable.

mathcass: I'll :+1: @ericboucher's comment to my comment and add that I gave a workaround here but can't recommend more getting more familiar with the CSV module. If you want any pointers or examples or anything from us, please let us know

Hagellach37: Hey,
I adjusted this and now we use the csv module. I agree, that this is cleaner. :)

transfer_results_module/transfer_results.py
Hagellach37: Hey,
when starting the script, the first thing that we do is to download the results from firebase and put everything into a `results.json` file. This file is deleted, when everything went fine at the end of the script. In case an error happens somewhere in between, the `results.json` file is not removed. 

When the script is started, it will check if there is a results.json file and if so, start to import the results that are stored there.

I'm not sure, if this already addressed the issue. If we reach a timeout, will the script not throw an error and stop at the position, where the error happened?

Hagellach37: Furthermore, we would not loose results, untill there are not deleted in firebase. When save_to_mysql fails then we will not delete the entries in firebase. The next time the script is started it will therefore also import these results again.

Hagellach37: But, I'm thinking on "what else could go wrong".

Our main problem would be if any results are lost. This means that results are not imported to the mysql, but deleted in firebase. "Old" results are lost after the "results.json" file is deleted. We have to make sure, that the file is not deleted if anything goes wrong.

mathcass: Sorry Benni, on a second look, I think you might have those bases covered. There are still some parts that make me a little uneasy, though, so I'd invite any comments or suggestions from @ericboucher if he has any

mathcass:
Is this still worth merging?

PR #17, opened 2018-03-15
Add waffle.io board

ericboucher:
I add the project to waffle.io. It is an extended ticket board working on top of GitHub issues and free for open source projects.
Let me know what you think!


mathcass:
I believe I've seen some of the HOTOSM projects use Waffle.io so let's give it a shot. :+1: from me and my vote is that we pick a timeframe (maybe 2 months) and try it out. We could have a brief async discussion about how to use it in the backend channel in Slack. 

Since @Hagellach37 has been doing most of the updates here, I added him as a reviewer and invite him to take a look at it. Benni, in a nutshell, it's a way to track bugs and feature requests tied to the repository. The advantage to this (rather than the GH board) is that it's a little more user friendly and gives you a better eye of what you should focus on.

mathcass:
Actually, after taking a look at waffle.io, I don't know if I can recommend it. It's requesting permissions on all organizations that I'm in (and I can't allow it to do that). You're welcome to use Waffle.io but I personally wouldn't be able to log in at all.

PR #19, opened 2018-05-15
Fix trailing comma in config

faevourite:
Hey there, MapSwipers. 👋Newbie here.

I'm just exploring the codebase and running the different scripts to get my feet wet before tackling any bigger tasks. I thought a good way to do that would be to contribute back smaller changes that I either had to make locally or that could help other new contributors get up-and-running faster.

In this PR, I'm just fixing a trailing comma in the sample config, which if used as a base for `config.cfg` will result in an error.

I was debating whether or not these small fixes are worth contributing one-by-one or in a big batch once I've gone through all the scripts. I figured sooner is better than later, but if you think doing it this way will result in too much overhead (review takes time, I know), please let me know and I'll close this and send a more complete PR later.

Thanks!


ericboucher:
Thanks for taking a stab at this @faevourite!

cfg/auth.py
```diff
@@ -70,20 +71,19 @@ def dev_firebase_admin_auth():
...
-        with open('../cfg/config.cfg') as json_data_file:
+        with open(CONFIG_PATH) as json_data_file:
             data = json.load(json_data_file)
             api_key = data['imagery'][tileserver]
             return api_key
-    except:
-        print("Something is wrong with your API key."
-              "Do you even have an API key?")
-        sys.exit()
+    except Exception:
```
mathcass: This part's a little tricky because there's possibly 3 different errors here. The `IOError` from the file open. The `ValueError` from the JSON parsing. And the `KeyError` from the dictionary access. 

One thing you could do is at the top level open the configuration and store it as a dictionary constant. So, at around line 12 have something like: 

```
try: 
    with open(CONFIG_PATH) as f:
        config_data = f.read()
except IOError:
    print('Unable to load configuration file at {}, exiting'.format(CONFIG_PATH))
    sys.exit(1)

try:
    CONFIG = json.loads(config_data)
except ValueError:
    print("Unable to parse configuration file, likely because of malformed JSON, exiting")
    sys.exit(1)
```

Then at the various places where we load it, we can just reference it as `CONFIG`, like 

```
api_key = CONFIG['imagery'][tileserver]
```

That'd be a good step toward DRYing up the code a little bit.

faevourite: Thanks, @mathcass . Applied your suggestion with a slight modification to use `raise` so that we'd get the stack trace printed out (the exit code is still 1).

PR #24, opened 2018-05-19
Fix imports for osgeo and utils

ericboucher:


import_module/run_import.py
```diff
@@ -11,15 +11,15 @@
 
 import logging
 import os
-import time
```
mathcass: So, it looks like you removed this import but it's used around this line, https://github.com/mapswipe/python-mapswipe-workers/pull/24/files#diff-37f15fe964b547fd83d5ac92f5118804R482

mathcass:
Can you update this import module? 

```
https://github.com/mapswipe/python-mapswipe-workers/blob/master/update_module/run_update.py#L9-L10
```

It should be 

```
from update_module.update_project_contributors import update_project_contributors
from update_module.update_project_progress import update_project_progress
```

mathcass:
Can you also change this line? 

* https://github.com/mapswipe/python-mapswipe-workers/blob/eric.p.boucher-fix-imports/transfer_results_module/transfer_results.py#L6

to `from utils import error_handling`?

cfg/auth.py
```diff
@@ -8,7 +8,7 @@
 import json
 import sys
 
-CONFIG_PATH = '../cfg/config.cfg'
+CONFIG_PATH = './cfg/config.cfg'
```
mathcass: This relative import isn't really working out correctly. This might do the trick though. I got it working locally: 

```
import os
dirname = os.path.dirname(__file__)
CONFIG_PATH = os.path.join(dirname, 'config.cfg')
```

Via, https://stackoverflow.com/questions/918154/relative-paths-in-python#918178

mathcass:
In addition to the last comment, could you go ahead and edit each instance of `sys.exit()` in `cfg/auth.py` to `sys.exit(1)`. 

Right now when we can't load the configuration as expected, we're exiting 0 when we should really be exiting 1.

cfg/auth.py
ericboucher: The issue is that if we install the files as a lib they actually shouldn't be in the same folder as `os.path.dirname(__file__)` which would be the installed library path, no? If the tests are running it means that this import is working in that case. Maybe we need to revise the script cwd?

mathcass: You're right, actually. They should be wherever they're installed. Mine was working because I had it installed via `python setup.py develop`. 

Something is amiss when we run this on the live server and I can't put my finger on it.

mathcass: Here it is, https://github.com/mapswipe/python-mapswipe-workers/blob/eric.p.boucher-fix-imports/cfg/loop_export_module.json#L4

Hagellach37: Hey,
sorry I got a bit out of touch with this discussion. 

On the server we're running the script in a virtual environment: `/data/environments/mapswipe_workers/bin/python3`

When installing the files as a lib in this virtual environment, we should be able to import them, not?

mathcass: @Hagellach37 Sorry to just now answer your question, yes, we should be able to import them. The problem ended up being that one of the pm2 configs hadn't been updated yet. Eric fixed that.

Hagellach37:
and now I tested it and found out what went wrong on the server. Your code @ericboucher worked fine, we just didn't adjust two lines in our config file on the server. Probably that's also the reason why it worked perfectly locally for @mathcass but not on the server. I will merge it now and then update everything on the server. :)

PR #27, opened 2018-05-26
Add error handling update modul

Hagellach37:
This pull request is adjusting the error handling in the update module. We didn't change this, as we did in the other module. Furthermore, I noticed that we import the slack client and the corresponding configuration separately from the other configuaritions, which are loaded in auth.py.
Due to this, we didn't adjust the path of the config file in `send_slack_messages.py`. This was the reason why no errors were reported in slack.

### DIFF
--- a/update_module/run_update.py
+++ b/update_module/run_update.py
@@ -10,7 +10,7 @@ from update_module.update_project_contributors import update_project_contributor
 from update_module.update_project_progress import update_project_progress
 
 from cfg.auth import firebase_admin_auth
-from utils.send_slack_message import send_slack_message
+from utils import error_handling
 
 import time
 import argparse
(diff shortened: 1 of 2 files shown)


utils/send_slack_message.py
```diff
@@ -3,13 +3,13 @@
 # Author: M. Reinmuth, B. Herfort
 ####################################################################################################
 import json
-
 from slackclient import SlackClient
 
+CONFIG_PATH = './cfg/config.cfg'
```
ericboucher: So actually I think this invites for a config file that loads the json from the CONFIG_PATH and exposes-it, and this file would import the config.py instead. See here as an example - https://github.com/BeaconLabs/encompass/blob/master/backend/config/config.py

Without getting into a file with that many functions, I think that at least loading the config in a config.py makes sense.

update_module/run_update.py
```diff
@@ -129,15 +129,8 @@ def run_update(project_selection, user_project_list, output_path):
...
-        except:
-            tb = sys.exc_info()
-            # log error
-            logging.warning(str(tb))
-            # send mail to mapswipe google group with
-            print(tb)
-            msg = str(tb)
-            head = 'google-mapswipe-workers: run_update.py: error occured'
-            send_slack_message(head + '\n' + msg)
+        except BaseException as error:
```
ericboucher: Why BaseException, can we not be a bit more precise here?

PR #29, opened 2018-09-27
Reduce group size to 40 tiles in width.

ericboucher:


import_module/create_groups.py
```diff
@@ -232,7 +232,7 @@ def get_vertical_slice(slice_infos, zoom):
     # the final group polygon is calculated from TileX_min, TileX_max, TileY_min, TileY_max
 
     # the width threshold defines how "long" the groups are
-    width_threshold = 70
+    width_threshold = 40
```
mathcass: So you have a rough idea of what this corresponds to in the app? Does this correspond with how many "swipes" there are (IE how many times you have to page through)? Or is it number of tiles wide? 

I'm just thinking about if there needs to be some sort of constraint on this for it to work out nicely. IE, does it need to be an even number because we show two tiles wide at a time. 

If we know the answer to that, it'd be a good comment or else point in the docs.

mathcass: Maybe @Hagellach37 knows?

PR #32, opened 2018-10-08
changes from master

Hagellach37:
There are some changes that have not been commited to the master related to the KML import issue Gabor had recently.

@ericboucher it seems that you changed something directly on the server. Is this correct?

And do we need all those print statements?

### DIFF
--- a/import_module/create_groups.py
+++ b/import_module/create_groups.py
@@ -449,15 +449,12 @@ def create_tasks(xmin, xmax, ymin, ymax, config):
 def create_groups(groups, config):
     # this function will create a json file to upload in firebase groups table
 
-    print(config)
+
     # Create the output Driver
     outDriver = ogr.GetDriverByName('GeoJSON')
-    print(outDriver)
     # Create the output GeoJSON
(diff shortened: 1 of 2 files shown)


import_module/create_groups.py
```diff
@@ -449,11 +449,15 @@ def create_tasks(xmin, xmax, ymin, ymax, config):
 def create_groups(groups, config):
     # this function will create a json file to upload in firebase groups table
 
+    print(config)
     # Create the output Driver
     outDriver = ogr.GetDriverByName('GeoJSON')
+    print(outDriver)
     # Create the output GeoJSON
-    outfile = 'data/groups_{}.geojson'.format(config["project_id"])
+    outfile = '/data/groups_{}.geojson'.format(config["project_id"])
```
ericboucher: @Hagellach37 this is actually the only line that mattered. It was a folder access issue. I would suggest checking paths and folder creations again or transitioning fully to docker maybe?

ericboucher: The outDataSource ended up being `None`

PR #128, opened 2019-06-21
feat (analytics): Track new user creation in Segment

asaschachar:
The following PR initializes firebase functions (for ovio-mapswipe-dev) and adds a firebase function `trackNewUser` to send an event to segment when new users are created.

Note that this PR does not move [existing firebase functions](https://github.com/mapswipe/mapswipe-firebase-functions) to this repo. It makes sense to consolidate all firebase functions and have them be close to the consumers of the backend APIs so you don't have to worry about deploying from two separate github repos if APIs change.

The write key of segment is stored in ovio-mapswipe-dev firebase as a configuration variable:
`firebase functions:config:set` to get the configuration variable. This write key is also available in the segment project [here](https://app.segment.com/mapswipe/sources/python_mapswipe_workers_firebase_functions/overview)

The deployed firebase function is [here](https://console.firebase.google.com/project/ovio-mapswipe-dev/functions/list)

cc: @ericboucher

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -1,0 +1,5 @@
+{
+  "projects": {
+    "default": "ovio-mapswipe-dev"
+  }
+}
--- a/firebase.json
+++ b/firebase.json
@@ -1,0 +1,7 @@
+{
+  "functions": {
+    "predeploy": [
+      "npm --prefix \"$RESOURCE_DIR\" run lint"
+    ]
+  }
+}
--- a/functions/.eslintrc.json
+++ b/functions/.eslintrc.json
@@ -1,0 +1,123 @@
+{
+  "parserOptions": {
+    // Required for certain syntax usages
+    "ecmaVersion": 2017
+  },
+  "plugins": [
+    "promise"
+  ],
+  "extends": "eslint:recommended",
+  "rules": {
+    // Removed rule "disallow the use of console" from recommended eslint rules
+    "no-console": "off",
+
+    // Removed rule "disallow multiple spaces in regular expressions" from recommended eslint rules
+    "no-regex-spaces": "off",
+
+    // Removed rule "disallow the use of debugger" from recommended eslint rules
+    "no-debugger": "off",
+
+    // Removed rule "disallow unused variables" from recommended eslint rules
+    "no-unused-vars": "off",
+
+    // Removed rule "disallow mixed spaces and tabs for indentation" from recommended eslint rules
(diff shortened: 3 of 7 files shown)


Matthias-Schaub:
Why do we need to track new user creation in Segment. What is Segment?
This would mean to use yet another thrid-party service right? 

Right now we are focusing on getting Mapswipe Workers 2.0 (+ change detection project type) stable and production ready and trying to not introduce new features.

On Firebase Functions having their own repository: The Firebase Functions have nothing to do with Python-Mapswipe-Workers. They only are deployed to the Firebase and are triggered by Firebase Realtime Database.
The Python Mapswipe Workers creates new projects with all their associated groups and task and pushes them to Firebase Realtime Database. It also will transfer the results to a Postgres database to relive (keep costs down) the Firebase Realtime Database.
At some point the documentation will be clear about the different parts of a working Mapswipe environment and on how to deploy them.

ericboucher:
@Matthias-Schaub Asa had most of the context. The idea behind segment was to try and create, then aggregate user and usage metrics to better understand user behavior on the app associated with data creation and results. It is an exploration at this point and maybe there are other tools that better suit your need for now, but Segment is quickly becoming an industry standard.

asaschachar:
@Matthias-Schaub Apologies for the delayed response on this one.

You don't have to track new user creation in Segment. Segment is nice because it works across many analytics providers, meaning you send data to Segment and then you can forward that data to the analytics provider of your choosing (ex: send it to Google Analytics, send it to Optimizely, or send it to both). We decided to use Segment here primarily because we didn't know what analytics provider you all would want the new user creation data to be in. Since Segment works across many different analytics providers, that is why it is becoming an industry standard to begin with as @ericboucher mentions above.

To add to @ericboucher's comment, the reason we wanted to track new user creation was because it was the first step in a larger project we were working on to provide some metrics by which mapswipe could better understand their user acquisition and behavior.

On repo organization: I agree that the firebase functions currently have nothing to do with the current Python Mapswipe Workers, so it makes sense to keep them in a separate repo. I guess my point was that if you do end up having the firebase functions provide APIs that are relied on consumers, it's easiest to manage compatibility of the provider and consumer of the APIs in the same repo. This is probably more likely to happen between the firebase functions and the mobile application now that I'm revisiting this.

Feel free to see this PR as purely an exploratory suggestion for how you might start tracking analytics data for analysis later. Thanks!

PR #214, opened 2019-10-16
Fix results key error

Hagellach37:
* check in mapswipe_workers to not transfer invalid results
* check in firebase functions, to not update counters when invalid results have been submitted


mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -92,9 +93,18 @@ def results_to_file(results, projectId):
     logger.info(f'Got %s groups for project {projectId} to transfer' % len(results.items()))
     for groupId, users in results.items():
         for userId, results in users.items():
-            timestamp = results['timestamp']
-            start_time = results['startTime']
-            end_time = results['endTime']
+
+            # check if all attributes are set, if not don't transfer the results for this group
```
ericboucher: Why would data not exist?

laurentS: We talked about it with Benni yesterday, after sentry spiked with 2.8k errors in under 24h, because of one result object that has no `result` attribute. We've not been able to understand why this happened, but it did happen only once in about 2 weeks of operation. The incoming task does not seem to be the issue, as at least one other user provided correctly formatted results for it. So I think that checking and logging on both sides (JS+python) is probably a good approach for now, as it might help identify where things go wrong, if it ever happens again.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -92,9 +93,18 @@ def results_to_file(results, projectId):
...
+            try:
+                timestamp = results['timestamp']
+                start_time = results['startTime']
+                end_time = results['endTime']
+                results = results['results']
+            except KeyError as e:
+                sentry.capture_exception_sentry(e)
```
laurentS: are the logs going to be available? (do they get exported from the docker container, etc...) if not, you could use sentry's message capture https://docs.sentry.io/error-reporting/capturing/?platform=python#capturing-messages to keep a record of what group/project/user this happened in.

mapswipe_workers/sample_data/test_invalid_results.json
```diff
@@ -0,0 +1,8 @@
+{
```
laurentS: Is there an actual test attached to this data, or is it just a record of what broken data looks like? It would be worth adding a comment in the JSON to explain what it means, otherwise it looks a lot like a random file committed by mistake :)

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: basicalll we do checks on JS to prevent the real time user stats to be wrong, and in python to prevent storing wrong data in the postgres database

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: a good idea to have this in sentry as well, will make it possible for us to track by just looking at sentry then.

mapswipe_workers/sample_data/test_invalid_results.json
Hagellach37: we don't have a test for this, I just used it to manually upload it to firebase to test it in dev? rather delete then?

Hagellach37:
@ericboucher 
the basic flow:
* a user uploads a result to firebase, which is somehow invalid because not all attributes are set
* once inserted into firebase this will trigger to update the progress of a project and user stats
* the back end will fetch data from firebase every 10 minutes, insert into postgres and after doing this successfully delete in firebase

mapswipe_workers/sample_data/test_invalid_results.json
laurentS: I would keep the file, it's useful. Maybe add a line in the json like below since json doesn't really support comments. Ideally explain briefly how to use it to test, because in 3 months, we'll all have forgotten:
```json
  // your other lines
{
  "comment": "this is for...."
}
```

Issue #212, opened 2019-10-16
KeyError  'results'


Hagellach37:
Make sure that we catch an error when the `results` key is not properly set in firebase results for a user and a group. E.g. like this:

```
    "g576" : {
      "aXBOSBgG1uMVADMfqAaWfQJe3U42" : {
        "endTime" : "2019-10-15T16:17:44.928Z",
        "startTime" : "2019-10-15T16:16:19.730Z",
        "timestamp" : "2019-10-15T16:17:44.928Z"
      }
    }
```

use try except KeyError statement and sentry.captureError.

Second, make sure that the firebase functions are not triggered when incomplete results are uploaded.

Issue #211, opened 2019-10-10
Generate Stats: Per Project Statistics for MapSwipe Analytics


Hagellach37:
# Per Project Statistics
For each project we will provide the following statistics and datasets

## By Date
Generate `[project_id]_by_date.json` , `[project_id]_by_date.csv` with the following attributes/columns:
* [x] project id
* [x] timestamp (e.g. on a daily basis like this `2019-09-29 00:00:00`, `2019-09-30 00:00:00`)
* [x] cumulative progress
* [x] progress (for this day)
* [x] cumulative number of results
* [x] number of results (number of results for this day)
* ~~[ ] cumulative number of tasks mapped (all tasks for which we got at least one result)~~
* ~~[ ] number of tasks mapped (all tasks for which we got at least one result for this day)~~
* ~~[ ] cumulative mapped area in square kilometers (the area of all tasks for which we got at least one result)~~
* ~~[ ] mapped area in square kilometers (the area of all tasks for which we got at least one result for this day)~~
* [x] cumulative number of active users / contributors (all users who contributed at least one result)
* [x] number of active users / contributors (all users who contributed at least one result for this day)

## Raw Aggregated Results
Generate `[project_id]_agg_results.geojson` and  `[project_id]_agg_results.csv` with the following attributes/columns:
* [x] project id
* [x] group id
* [x] task id
* [x] number of results
* [x] number of `0` results
* [x] number of `1` results
* [x] number of `2 results
* [x] number of `3` results
* [x] share of `0` results
* [x] share of `1` results
* [x] share of `2 results
* [x] share of `3` results
* [x] agreement
* [x] geometry

Issue #210, opened 2019-10-10
Generate Stats: Project Statistics for MapSwipe Analytics


Hagellach37:
# Overall Project Statistics
Generate ~~`projects.json`~~, `projects_geom.geojson`, ` projects_centroid.geojson` and `projects.csv` with the following attributes/columns:

## Static
* [x] id
* [x] name
* [x] description
* [x] look_for
* [x] type
* ~~[ ] number of groups~~
* ~~[ ] number of tasks~~
* [x] geometry
* [x] centroid
* [x] area in square kilometers
* [x] status (e.g. active, inactive, finished)

## Dynamic
* [x] id
* [x] progress
* [x] number of results
* [x] number of results progress (only results which contribute to progress)
* ~~[ ] number of tasks mapped (all tasks for which we got at least one result)~~
* ~~[ ] mapped area in square kilometers (the area of all tasks for which we got at least one result)~~
* [x] number of active users / contributors (all users who contributed at least one result)
* [x] day (last time the data has been updated)

Issue #205, opened 2019-10-08
Generate Stats: improve performance


Hagellach37:
Currently some parts of `generate stats` just take to long. This is also related to #203 and we need to think about which numbers we want to provide.

| dataset | currently | optimized |
| -- | -- | -- |
| agg_results | ~45min | |
| agg_res_by_user_id | ~35min | |
| agg_res_by_project_id | ~35min | |
| agg_res_by_project_id_geom | ~10min | |
| agg_res_by_project_id_centroid | ~30min | |
| agg_projects | <1sec | |
| agg_projects_by_project_type | <1sec | |
| agg_users | ~10min | |
| agg_progress_by_project_id | ~12min | |
| agg_progress_by_project_id_geom | ~12min | |
| agg_progress_by_project_id_centroid | ~15min | |
| agg_res_by_task_id (small project, or few results) | ~1sec | |
| agg_progress_by_project_id_and_date | ~1sec | |
| agg_res_by_user_id_and_date | <1sec | |
 
In general queries that are based on `aggregated_results` seem to take longer...

### Immediate tasks:
* [x] use only 1 query for `agg_res_by_project_id`, `agg_res_by_project_id_geom` and `agg_res_by_project_id_centroid` instead of 3
* [x] use only 1 query for `agg_progress_by_project_id`. `agg_progress_by_project_id_geom` and `agg_progress_by_project_id_centroid`

PR #218, opened 2019-10-30
Generate statistics for projects

Hagellach37:
* improve statistics generatation #205 
* generate per project statistics #211 
* generate project statistics #210 

further fixes:
* results key error #212
* capture messages using sentry #214


postgres/scripts/dbdump.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-DBUSER="backup"
-DBNAME="mapswipe"
-DBPASSWD="backupuserpassword"
+SSH_REMOTE_HOST=username@111.11.111.0
+PGUSER="backup"
+PGDATABASE="mapswipe"
+PGPASSWORD="backupuserpassword"
```
laurentS: if these are placeholders, why not, but otherwise, you shouldn't commit passwords

mapswipe_workers/Dockerfile
```diff
@@ -12,10 +12,12 @@ RUN mkdir -p /usr/local/mapswipe_workers/
...
-RUN mkdir -p /var/lib/mapswipe_workers/api-data/agg_res_by_task_id/
-RUN mkdir -p /var/lib/mapswipe_workers/api-data/agg_res_by_project_id_and_date/
-RUN mkdir -p /var/lib/mapswipe_workers/api-data/agg_progress_by_project_id_and_date/
-RUN mkdir -p /var/lib/mapswipe_workers/api-data/agg_res_by_user_id_and_date/
+RUN mkdir -p /var/lib/mapswipe_workers/api-data/agg_results/
+RUN mkdir -p /var/lib/mapswipe_workers/api-data/groups/
+RUN mkdir -p /var/lib/mapswipe_workers/api-data/history/
+RUN mkdir -p /var/lib/mapswipe_workers/api-data/projects/
+RUN mkdir -p /var/lib/mapswipe_workers/api-data/results/
+RUN mkdir -p /var/lib/mapswipe_workers/api-data/tasks/
```
laurentS: I would try to use a variable for these repeated paths, it helps avoid typos, and also helps confirm that you indeed want `/var/log` and not `/var/lib` on the first line, it helps confirm that the line below should indeed by `/usr/local`, and so on.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -201,4 +208,20 @@ def get_user_ids_from_results(results):
...
+    '''
+
+    pg_db = auth.postgresDB()
+    sql_query = '''
+        SELECT project_id from projects;
+    '''
+    raw_ids = pg_db.retr_query(sql_query, None)
+    project_ids = [i[0] for i in raw_ids]
+
+    del pg_db
```
laurentS: Is there a particular reason to use `del` here? My understanding is that since `pg_db` is a local variable, it will get deleted at the end of the function, but please correct me if I'm wrong. See [this SO answer](https://stackoverflow.com/a/2193071/1138710) for reference. If it is indeed useful to have the `del` statement here, maybe add a comment to explain why?

mapswipe_workers/mapswipe_workers/firebase_to_postgres/update_data.py
```diff
@@ -70,7 +70,7 @@ def update_user_data(user_ids=None):
             except KeyError:
                 # if user has no "username" attribute, we set it to None
                 username = None
-                logger.info(f"user {user_id} didn't have a created attribute set it to {username}")
+                logger.info(f"user {user_id} didn't have a username attribute set it to {username}")
```
laurentS: I'm not sure I understand this logged comment, can you rephrase it to be clearer?

mapswipe_workers/mapswipe_workers/generate_stats/generate_stats.py
```diff
@@ -0,0 +1,46 @@
+from mapswipe_workers.definitions import logger
+from mapswipe_workers.definitions import DATA_PATH
+
+from mapswipe_workers.generate_stats import project_stats
+from mapswipe_workers.generate_stats import overall_stats
```
laurentS: See my other note, you can merge these `import` lines into a single one, it's easier to read :)

mapswipe_workers/mapswipe_workers/generate_stats/generate_stats.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def generate_stats(project_id_list, all_projects=False):
```
laurentS: I would add a docstring to this function, to make it clear what it does. The one in `transfer_results` was pretty good.

mapswipe_workers/mapswipe_workers/generate_stats/generate_stats.py
```diff
@@ -0,0 +1,46 @@
...
+
+    projects_info_filename = f'{DATA_PATH}/api-data/projects/projects_static.csv'
+    projects_df = overall_stats.get_project_static_info(projects_info_filename)
+
+    projects_info_dynamic_filename = f'{DATA_PATH}/api-data/projects/projects_dynamic.csv'
+    projects_dynamic_df = overall_stats.load_project_info_dynamic(projects_info_dynamic_filename)
+
+    if all_projects:
```
laurentS: Is it correct that you either call `generate_stats([some_list])` or `generate_stats(??, True)`? If so, I would suggest a bit of defensive programming at the start of the function, to check that the args make sense, maybe `if project_id_list is not None: assert all_projects == False` etc... This would help catch bugs where a dev calls the function the wrong way.
Or why not write a wrapper around `generate_stats(project_id_list)` like `generate_stats_for_all_projects()` in which you just get the complete id list like you do here on line 17, and then call `generate_stats(the_full_list)`? I think this would make the intent clearer elsewhere in the codebase.

mapswipe_workers/mapswipe_workers/generate_stats/generate_stats.py
```diff
@@ -0,0 +1,46 @@
...
+        project_id_list = projects_df['project_id'].to_list()
+        logger.info(f'will generate stats for all projects.')
+
+    logger.info(f'will generate stats for: {project_id_list}')
+
+    # get per project stats and aggregate based on task_id
+    for project_id in project_id_list:
+
+        # check if project id is correct
+        if not project_id in projects_df['project_id'].to_list():
```
laurentS: It would probably make sense to factor the `projects_df['project_id'].to_list()` out of the for loop, so you don't run it `len(project_id_list)` times, as it's probably quite expensive.

mapswipe_workers/mapswipe_workers/generate_stats/generate_stats.py
```diff
@@ -0,0 +1,46 @@
...
+        logger.info(f'start generate stats for project: {project_id}')
+        idx = projects_dynamic_df.index[projects_dynamic_df['project_id'] == project_id].tolist()
+        if len(idx) > 0:
+            projects_dynamic_df.drop([idx[0]], inplace=True)
+
+        # aggregate results and get per project statistics
+        project_stats_dict = project_stats.get_per_project_statistics(project_id)
+        if project_stats_dict:
+            projects_dynamic_df = projects_dynamic_df.append(project_stats_dict, ignore_index=True)
+            projects_dynamic_df.to_csv(projects_info_dynamic_filename, index_label='idx')
```
laurentS: you're overwriting this CSV for each `project_id` in the loop. Is this what you want?

mapswipe_workers/mapswipe_workers/generate_stats/overall_stats.py
```diff
@@ -0,0 +1,53 @@
...
+            FROM projects
+        ) TO STDOUT WITH CSV HEADER"""
+
+    with open(filename, 'w') as f:
+        pg_db.copy_expert(sql_query, f)
+
+    del pg_db
+    logger.info('got projects from postgres.')
+
+    df = pd.read_csv(filename)
```
laurentS: Could it be more efficient to read the dataframe directly from the DB with [pd.read_sql_query](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_sql_query.html#pandas.read_sql_query) ?

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
```diff
@@ -0,0 +1,345 @@
...
+    sql_query = sql.SQL(
+        'COPY (SELECT * FROM results WHERE project_id = {}) TO STDOUT WITH CSV HEADER'
+    ).format(sql.Literal(project_id))
+
+    with open(filename, 'w') as f:
+        pg_db.copy_expert(sql_query, f)
+    del pg_db
+    logger.info(f'got results from postgres for {project_id}')
+
+    df = pd.read_csv(filename)
```
laurentS: I would try to refactor this code bit to load a DataFrame from a CSV query into a function. You have it in at least 3 different places, so you could do something like:
```python
def load_df_from_sql(query):
   # your code above, or better, using pandas.read_sql_query
   return df
```

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
```diff
@@ -0,0 +1,345 @@
...
+        results_by_task_id_df[1] = 0
+
+    if 2 not in results_by_task_id_df.columns:
+        results_by_task_id_df[2] = 0
+
+    if 3 not in results_by_task_id_df.columns:
+        results_by_task_id_df[3] = 0
+
+    results_by_task_id_df['total_count'] = results_by_task_id_df[0] + results_by_task_id_df[1] + results_by_task_id_df[2] + results_by_task_id_df[3]
+    results_by_task_id_df['0_share'] = results_by_task_id_df[0] / results_by_task_id_df['total_count']
```
laurentS: what happens if `total_count == 0` for a row?

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
```diff
@@ -0,0 +1,345 @@
...
+    progress_by_date_df['cum_number_of_results'] = progress_by_date_df['number_of_results'].cumsum()
+    progress_by_date_df['cum_number_of_results_progress'] = progress_by_date_df['number_of_results_progress'].cumsum()
+    progress_by_date_df['progress'] = progress_by_date_df['number_of_results_progress'] / required_results
+    progress_by_date_df['cum_progress'] = progress_by_date_df['cum_number_of_results_progress'] / required_results
+
+    logger.info('calculated progress by date')
+    return progress_by_date_df
+
+
+def get_new_user(day, first_day):
```
laurentS: I'm probably splitting hairs here, but I would rename this to `is_new_user` because `get_new_user` makes me think the function would return a user, not a boolean.

mapswipe_workers/mapswipe_workers/generate_stats/user_stats.py
```diff
@@ -0,0 +1,44 @@
+def get_aggregated_results_by_user_id(filename):
```
laurentS: Is this function expected to not return anything?

mapswipe_workers/mapswipe_workers/utils/geojson_functions.py
```diff
@@ -0,0 +1,62 @@
...
+    '''
+    Go through geojson file and try to cast all values as float, except project_id
+    remove redundant geometry property
+    '''
+
+    filename = filename.replace('csv', 'geojson')
+    with open(filename) as f:
+        geojson_data = json.load(f)
+
+    if len(geojson_data['features']) > 0:
```
laurentS: What happens if `features` is empty? is it an error worth logging or reporting?

Issue #234, opened 2019-11-25
Firebase Transfer Results: This request would cause too many functions to be triggered.


Hagellach37:
Since we are using Firebase Cloud Functions and Triggers, this has a consequence when we want to delete many entries in Firebase after transferring results to postgres. The effect of this is, that after the data has been stored in postgres, the delete in firebase fails. Consequently, we'll try to import the very same data from firebase in our next iteration. More information on limits of firebase database here: https://firebase.google.com/docs/database/usage/limits

We need to find a way to delete the data in firebase properly, once all results have been transferred to postgres.

we need to thing about either limit the number of groups that should be transferred (and deleted later). E.g. as described here: https://firebase.google.com/docs/database/admin/retrieve-data#limit-queries

Or we use a different type of triggers. For example such as `onCreate` or `onUpdate` instead of `OnWrite`. E.g. as described here: https://firebase.google.com/docs/functions/database-events

When trying to delete the entries in firebase we get the following response:

412 Precondition Failed | The request's specified ETag value in the if-match header did not match the server's value.
-- | --

This is due to the fact, that the insert took so long, and the REST api connection has been resetting, and firebase is in a new state then.

Hagellach37:
This has been partly fixed by using `OnCreate` for the firebase cloud fucntion trigger. 
The problem related to performance and resetting https connection will be addressed in another issue.

PR #235, opened 2019-11-25
use onCreate for counter function #234

Matthias-Schaub:


firebase/functions/index.js
```diff
@@ -8,16 +8,9 @@ admin.initializeApp()
...
-    // (e.g. when deletion during transfer_results() occured)
-    // do nothing
-    if (!snapshot.after.exists()) {
-        return null
-    }
-
+exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{userId}/').onCreate((snapshot, context) => {
     const promises = []
-    const result = snapshot.after.val()
+    const result = snapshot.val()
```
Hagellach37: there is no `after.val()` for onCreate, right?

Hagellach37:
make sure to test this in dev before merging the hotfix into master.

PR #243, opened 2019-11-28
Feature new imagery sources

Hagellach37:


manager_dashboard/manager_dashboard/js/forms.js
```diff
@@ -30,6 +30,23 @@ function displayProjectTypeFormular(projectType) {
...
+        document.getElementById("tileServerCredits"+projectType+which).value = "© 2019 Maxar"
+    } else if (tileServer == 'esri' | tileServer == 'esri_world') {
+        document.getElementById("tileServerCredits"+projectType+which).value = "© 2019 ESRI"
+    } else if (tileServer == 'mapbox') {
+        document.getElementById("tileServerCredits"+projectType+which).value = "© 2019 MapBox"
+    } else if (tileServer == 'sinergise') {
+        document.getElementById("tileServerCredits"+projectType+which).value = "© 2019 Sinergise"
+    } else if (tileServer == 'custom') {
+        document.getElementById("tileServerCredits"+projectType+which).value = "Please add imagery credits here."
+    }
```
laurentS: This works, but you can make it easier to read with something like a [switch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) or using a mapping like this:
```javascript
credits = {
"bing": "bing credit text",
"esri": "esri credit text"
// ... the other ones
};
document.getElementById("tileServerCredits"+projectType+which).value = credits[tileServer];
```
It works well here because you only have a handful of fixed values. It tends to be much more maintainable, because you write everything just once.

mapswipe_workers/mapswipe_workers/project_types/change_detection/tile_functions.py
```diff
@@ -57,8 +56,11 @@ def lat_long_zoom_to_pixel_coords(lat, lon, zoom):
     p = Point()
     sinLat = math.sin(lat * math.pi / 180.0)
     x = ((lon + 180) / 360) * 256 * math.pow(2, zoom)
-    y = (0.5 - math.log((1 + sinLat) / (1 - sinLat)) /
-         (4 * math.pi)) * 256 * math.pow(2, zoom)
+    y = (
```
laurentS: it looks like this file or most of it is duplicated between the different project types. Do you actually need that? Is there a way you can refactor these functions/classes into a file that you can then import in the various project types?

Hagellach37: Totally true. I moved the tile functions to `utils`.

manager_dashboard/manager_dashboard/js/forms.js
Hagellach37: I like your approach. It is much simpler.

laurentS: yes, it does look a lot cleaner now!

laurentS:
A quick suggestion about your last commit: I would tend to separate linting commits like when you run `black` in a separate commit from actual code changes. usually I commit something like `Run black on file xxx.py` so I know this commit basically doesn't change anything, and doesn't need to be reviewed. And then I commit my actual changes, so I don't have pages and pages of linting details drowning out the useful stuff. Not a disaster, but it makes finding errors much easier in the pile of commits :)

Hagellach37:
alright. We are using pre-commit now. So basically whenever we touch an "old" file and change something in the code this will also trigger black. Maybe one option to prevent this in the future would be to run black on all files once. From then on the changes should only reflect the actual edits to the code and not formatting related things.

PR #250, opened 2019-12-18
v2.1.7

Hagellach37:
* [x]  fix max zoom level
* [x] new tutorial for change detection
* [x] fix small bug in creation of built area tutorial


mapswipe_workers/mapswipe_workers/project_types/change_detection/change_detection_project.py
```diff
@@ -117,7 +117,13 @@ def validate_geometries(self):
 
             # calculate max area based on zoom level
             # for zoom level 18 this will be 5000 square kilometers
-            max_area = (20 - int(self.zoomLevel)) * (20 - int(self.zoomLevel)) * 1250
+            # max zoom level is 22
```
laurentS: this `validate_geometries` function looks like it's very similar to the one in `build_area_project.py`. Is there a way to refactor them? Maybe as a method on `BaseProject` ? and if there are only small differences, you could call `self.the_bit_that_is_different()` from `BaseProject.validate_geometries()` which would then go into the subclass. I think you would end up with a lot less code, just by looking at it quickly.

mapswipe_workers/sample_data/build_area_tutorial.json
```diff
@@ -10,7 +10,8 @@
   "contributorCount": 0,
   "examplesFile": "sample_data/build_area_tutorial_tasks.csv",
   "tileServer":{
-    "name": "bing"
+    "name": "bing",
+    "apiKey": ""
```
laurentS: a small trick when writing dicts like this, add a colon to every line, even the last one, because it helps in keeping your diffs smaller, eg: if you had `"name": "bing",` (not the comma at the end) then the diff would have only been a single line here, making it more obvious that you just added the line with `apiKey`. It's details, but when you have big commits, these small bits do add up to a lot more to review. I think `black` does this for you, actually.

mapswipe_workers/sample_data/change_detection_tutorial_bing.json
```diff
@@ -0,0 +1,60 @@
...
+      "post_correct": "Yes, that looks like a new settlement being developed. Plenty of new houses and huts have been built in this area. Swipe left to continue the tutorial.",
+      "post_wrong": "Oh no. Make sure to spot all the new settlements and swipe right. New roads can also function as an indicator sometimes."
+    },
+    "changes_2": {
+      "pre": "And again: If new settlements have been built, *SWIPE* to the right side.",
+      "post_correct": "Correct. We can spot new settlements in the upper right part of the satellite image. Swipe left to go on with some more examples.",
+      "post_wrong": "Take a closer look. There are new settlements here in the upper right part of the satellite image."
+    },
+    "no_changes_1": {
+      "pre": "In this example, there are many buildings but *NO NEW SETTLEMENTS* emerged. *SWIPE* left to indicate no changes.",
```
laurentS: I think it's worth adding a step where buildings have disappeared, to say they do count as changes (unless this is a per project decision?). It was one of the first questions I got during user testing yesterday

mapswipe_workers/sample_data/build_area_tutorial.json
Hagellach37: Hey,
I agree, but I think here it's not going to work since this is inside the JSON file. But a good comment for the other parts of the code. :)

mapswipe_workers/mapswipe_workers/project_types/change_detection/change_detection_project.py
Hagellach37: Good comment. I will add a separate issue for this, because it will also have effects for the footprint project type. Currently we define this function for each project type. But a lot of it is redundant.

mapswipe_workers/sample_data/build_area_tutorial.json
laurentS: wooops, you're right, I commented too quickly there, sorry :)

mapswipe_workers/mapswipe_workers/project_types/change_detection/change_detection_project.py
Hagellach37: created new issue #253

PR #252, opened 2019-12-18
Feature tasking manager geometries

Hagellach37:


mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -29,7 +29,8 @@ class PythonLiteralOption(click.Option):
     def type_cast_value(self, ctx, value):
         try:
             return ast.literal_eval(value)
-        except:
+        except Exception as e:
```
laurentS: This is a "catch all" `except`, which you should avoid, because it can hide unexpected errors. In theory, you use an `except` to catch errors that you know might happen, and handle them in a specific way.
When you catch any exception like this (all exceptions in python are subclasses of `Exception`, so you get everything here), something you hadn't expected might happen, and you won't know about it.
The better way to do this is to limit your handling to `except TheExceptionYouKnowMayHappen as e:` and if something else happens, you will then get a clear message. https://stackoverflow.com/questions/14797375/should-i-always-specify-an-exception-type-in-except-statements for some more explanations

mapswipe_workers/mapswipe_workers/utils/geojson_functions.py
```diff
@@ -95,3 +97,105 @@ def add_metadata_to_geojson(filename: str, geometry_field: str = "geom"):
...
+            feature = ogr.Feature(featureDefn)
+            # create polygon from wkt and set geometry
+            feature.SetGeometry(geom)
+            # set other attributes
+            feature.SetField("id", counter)
+            # add feature to layer
+            layer.CreateFeature(feature)
+
+    layer = None
+    logger.info("created outifle: %s." % outfile)
```
laurentS: typo `outifle` -> `outfile`

mapswipe_workers/mapswipe_workers/utils/geojson_functions.py
```diff
@@ -95,3 +97,105 @@ def add_metadata_to_geojson(filename: str, geometry_field: str = "geom"):
...
+
+    # create fields
+    field_id = ogr.FieldDefn("id", ogr.OFTString)
+    layer.CreateField(field_id)
+
+    counter = 0
+    if not geometries:
+        logger.info("there are no geometries to save")
+    else:
+        for geom in geometries:
```
laurentS: in python you can do `for counter, geom in enumerate(geometries):` to get the same result in a cleaner way.

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
Hagellach37: You're right and got me... Need to find out which error will happen for this case. :D

Hagellach37:
Hey @laurentS,
thanks for the comments. I adjusted the bits.

PR #261, opened 2020-01-13
Add project types as global var to definitions.

Matthias-Schaub:
- Add project types as global var to definitions.
- Run Black.
- Make use of PROJECT_TYPES in mapswipe_workers.py
- Remove not needed imports in definitions.py and mapswipe_workers.py

### DIFF
--- a/mapswipe_workers/mapswipe_workers/definitions.py
+++ b/mapswipe_workers/mapswipe_workers/definitions.py
@@ -1,40 +1,41 @@
-import os
-import json
 import logging
 import logging.config
-from logging.handlers import TimedRotatingFileHandler
+import os
 
+from mapswipe_workers.project_types.build_area.build_area_project import (
+    BuildAreaProject,
+)
+from mapswipe_workers.project_types.change_detection.change_detection_project import (
+    ChangeDetectionProject,
+)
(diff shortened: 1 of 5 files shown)


mapswipe_workers/mapswipe_workers/definitions.py
```diff
@@ -1,40 +1,41 @@
...
+DATA_PATH = os.path.abspath("/var/lib/mapswipe_workers/")
+
+PROJECT_TYPE_CLASSES = {
+    1: BuildAreaProject,
+    2: FootprintProject,
+    3: ChangeDetectionProject,
+}
+
+PROJECT_TYPE_NAMES = {
+    1: BuildAreaProject.name,
```
Hagellach37: make sure to use the correct attribute name here. For BuiltAreaProject class you use `ProjectTypeName` instead of `name`.

mapswipe_workers/mapswipe_workers/project_types/change_detection/change_detection_project.py
```diff
@@ -19,6 +19,7 @@ class ChangeDetectionProject(BaseProject):
     """
 
     projectType = 1
```
Hagellach37: the projectType for Change Detection should be `3`and not `1`.

PR #263, opened 2020-01-13
Archive feature

Matthias-Schaub:


mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
```diff
@@ -0,0 +1,41 @@
...
+from mapswipe_workers import auth
+from mapswipe_workers.definitions import logger
+
+
+def archive_project(project_id: str) -> None:
+    """
+    Archive a project.
+
+    Call a function which deletes groups, tasks and results in Firebase.
+    Set archive = true for project in Postgres.
```
Hagellach37: minor change: you change to status = active

mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
```diff
@@ -0,0 +1,41 @@
...
+    ref.set({"archived"})
+
+    pg_db = auth.postgresDB()
+    sql_query = "UPDATE projects SET status = 'archived' WHERE project_id = {0}".format(
+        project_id
+    )
+    pg_db.query(sql_query)
+
+
+def delete_project_from_firebase(project_id: str) -> None:
```
Hagellach37: I find the function name a bit counter intuitive since only groups, tasks and results are deleted. maybe we can use `archive_project_in_firebase` or another term here

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -288,6 +294,21 @@ def run_create_tutorial(input_file):
         logger.exception(e)
 
 
+@click.command("archive")
+@click.option(
```
Hagellach37: we could use a similar approach here as in `run_generate_stats`. There we have `project_id_list` as an input.

mapswipe_workers/tests/test_archive_project.py
```diff
@@ -0,0 +1,50 @@
...
+        """Set up before tests are run."""
+        cls.project_id = "test_archive_project"
+        # TODO: Create project
+        archive_project.archive_project(cls.project_id)
+
+    @classmethod
+    def tearDownClass(cls):
+        """Tear down after tests are run."""
+        pg_db = auth.postgresDB()
+        sql_query = "".format(cls.project_id)
```
Hagellach37: I'm not sure what happens at this step or what is intended.

mapswipe_workers/tests/test_archive_project.py
```diff
@@ -0,0 +1,50 @@
...
+        pg_db.query(sql_query)
+
+        fb_db = auth.firebaseDB()
+        ref = fb_db.reference("v2/projects/{0}".format(cls.project_id))
+        ref.set({})
+
+    def test_firebase_changes(self):
```
Hagellach37: I think we should also test if the status of the project is actually set to `archived`.

postgres/initdb.sql
```diff
@@ -2,7 +2,6 @@
 CREATE EXTENSION postgis;
 
 CREATE TABLE IF NOT EXISTS projects (
-    archive boolean,
```
Hagellach37: what is the consequence of this once we deploy it? Initdb.sql will not be run when there is a database already. So we need to manually remove this column, right?

mapswipe_workers/tests/test_archive_project.py
```diff
@@ -0,0 +1,50 @@
...
+        ref = fb_db.reference("v2/tasks/{0}".format(self.project_id))
+        self.assertFalse(ref.get())
+
+        ref = fb_db.reference("v2/results/{0}".format(self.project_id))
+        self.assertFalse(ref.get())
+
+    def test_postgres_changes(self):
+        """Test if postgres project is archived."""
+        pg_db = auth.postgresDB()
+        sql_query = "SELECT archived FROM projects WHERE project_id = {}".format(
```
Hagellach37: you need to select the `status` here instead of `archived`. The `archived` attribute is not used anymore.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
Matthias-Schaub: Will remove the function and implement functionality in archive function.

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
Matthias-Schaub: Will add support for lists as argument.

mapswipe_workers/tests/test_archive_project.py
Matthias-Schaub: Proper tests will be implemented in the unittest branch.

mapswipe_workers/tests/test_archive_project.py
Matthias-Schaub: Proper tests will be implemented in the unittest branch.

postgres/initdb.sql
Matthias-Schaub: Yes.

PR #265, opened 2020-01-14
Rewrite of the cli mapswipe_workers.py.

Matthias-Schaub:
- Scheduling is now only used for run() (create projects, transfer results and generate stats).
- Code has been simplefied.
- Black and flake8 compatible.

### DIFF
--- a/mapswipe_workers/setup.py
+++ b/mapswipe_workers/setup.py
@@ -1,22 +1,23 @@
 from setuptools import setup, find_packages
 
 
-console_scripts='''
+console_scripts = """
     [console_scripts]
     mapswipe_workers=mapswipe_workers.mapswipe_workers:cli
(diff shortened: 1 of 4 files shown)


mapswipe_workers/setup.py
```diff
@@ -1,22 +1,23 @@
...
 setup(
-    name='mapswipe-workers',
-    version='3.0',
-    description='Install script for the mapswipe Python workers.',
-    author='B. Herfort, M. Schaub, M. Reinmuth',
-    author_email='',
-    url='www.mapswipe.org',
-    packages=find_packages(exclude=('tests', 'docs')),
+    name="mapswipe-workers",
+    version="3.0",
```
Hagellach37: should we correct this version to the actual version?

Issue #199, opened 2019-10-01
Only useful messages in Slack


Hagellach37:
Only send mesages useful for project managers in slack. All the other messages, and error logs should be going to sentry or into the logs only.

In slack we want:
* [x] successfull project creation
* [x] failed project creation
* [x] ~project progress reached 90%~
* [x] ~project progress reached 100%~

Also regarding the slack token:
* [x] make sure to not use the "benni" slack token

Matthias-Schaub:
To keep things simply Sentry and Logs should contain the same information.
Sentry (sentry_sdk.init()) should already recive all logging events with level error including previous logs at info level or high as "breadcrumbs" attached to it.
I propose to take another look into this to make sure logs from the logging module are also send to sentry. This way there is no need to consult to logging systems in case one must do so.

Another proposal would be to send Slack messages only from the cli module (mapswipe_workers.py). From there after success or failure of a major routine of the program (like add a new project) information can be sent to project administrators and others in Slack. This would make the program more consise.

PR #267, opened 2020-01-15
Logging, Sentry and Slack

Matthias-Schaub:
Refactored the sentry and slack.
This is related to #199.


mapswipe_workers/mapswipe_workers/utils/slack_helper.py
```diff
@@ -0,0 +1,34 @@
...
+        message = (
+            "### PROJECT CREATION FAILED ###\n"
+            + "Project Name: {0}\n".format(project_name)
+            + "Project draft is deleted."
+        )
+    else:
+        # TODO: Raise an Exception
+        pass
+
+    slack_client.chat_postMessage(channel=slack_channel, text=message)
```
Hagellach37: just for clarification: we don't need the username anymore, right?

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -181,20 +159,15 @@ def run_create_tutorial(input_file) -> None:
...
 @click.command("archive")
 @click.option(
-    "--project-id",
-    "-i",
-    help=("Archive project with giving project id"),
-    type=str,
+    "--project-id", "-i", help=("Archive project with giving project id"), type=str,
 )
 @click.option(
     "--project-ids",
```
Hagellach37: I'm a bit confused that we have both options (project-id and project-ids). But I think this is fine, just need to tell users how to use it properly in the docs.

mapswipe_workers/mapswipe_workers/utils/slack_helper.py
Matthias-Schaub: As far as I understand the user parameter is only used for sending an ephemeral message: "Ephemeral messages are only visible to an assigned user in a specified channel, is nearly the same as sending a regular message, but with an additional user parameter." https://slack.dev/python-slackclient/basic_usage.html

So yes, we do not nee the username anymore

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
Matthias-Schaub: I have let this option in the code because I find the list option kind of unintuitive. If one wants to archive only one project one has to write: `mapswipe_workers archive --project-ids ["my_project_id"]` instead of simply ``mapsiwpe_workers archive --project_id my_project_id`.

Maby it is possible to have on option with two types of input. Either a list or simply a string.

Should we open an issue for that? This would apply also to the other cli arguments such as transfer results etc.

Hagellach37: let's keep it as you did. I think we don't need another issue for this.

PR #269, opened 2020-01-15
Travis template. Run Flake8 and Black. Install mapswipe_workers.

Matthias-Schaub:


.travis
```diff
@@ -0,0 +1,19 @@
+# This is a special configuration file to run tests on Travis-CI via
+# GitHub notifications when changes are committed.
+#
+# For technical details, see http://travis-ci.org/
+
+language: python
+python:
+  - "3.6"
```
Hagellach37: do we use string formatting with f'fdafd.{test}... This works from python 3.7 on, right?

Matthias-Schaub: Good point. Have not thought about it. But f-strings are introduced in Python 3.6.

PR #272, opened 2020-01-27
Remove 'archived' attribute. 'archived' is now a value of status attr…

Matthias-Schaub:
…ibute.


mapswipe_workers/mapswipe_workers/base/base_project.py
```diff
@@ -221,7 +220,6 @@ def save_to_postgres(self, project, groups, groupsOfTasks):
             '''
```
Hagellach37: we also need to remove one `%s` in the line above.

Matthias-Schaub: Removed placeholder. Good catch!

PR #281, opened 2020-02-02
add script to archive old firebase projects

Hagellach37:
* Run a script to move the old projects (created before we switched to v2) to the new Firebase projects path.
* Delete the old groups
* Delete old results, imports, announcements
* Don't touch old user data

This workflow would be run once for the production environment. I'm not sure where is the best place to put it. I put it to a new directory `scripts` for now.

### DIFF
--- a/mapswipe_workers/scripts/archive_old_projects.py
+++ b/mapswipe_workers/scripts/archive_old_projects.py
@@ -14,16 +14,32 @@ def get_old_projects():
     return projects
 
 
-def move_project_data_to_v2(project_id, project_data):
+def move_project_data_to_v2(project_id):
     """
     Copy project information from old path to v2/projects in Firebase.
     Add status=archived attribute.
+    Use Firebase transaction function for this.
     """
-    project_data["status"] = "archived"
+
+    # Firebase transaction function
+    def transfer(current_data):
+        current_data["status"] = "archived"
+        current_data["projectType"] = 1
+        fb_db.reference("v2/projects/{0}".format(project_id)).set(current_data)
+        return dict()
+
     fb_db = auth.firebaseDB()
-    fb_db.reference("v2/projects/{0}".format(project_id)).set(project_data)
-    fb_db.reference("projects/{0}".format(project_id)).set({})
-    logger.info(f"moved old project to v2: {project_id}")
+    projects_ref = fb_db.reference(f"projects/{project_id}")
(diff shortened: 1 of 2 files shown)


mapswipe_workers/scripts/archive_old_projects.py
```diff
@@ -0,0 +1,69 @@
...
+
+def move_project_data_to_v2(project_id, project_data):
+    """
+    Copy project information from old path to v2/projects in Firebase.
+    Add status=archived attribute.
+    """
+    project_data["status"] = "archived"
+    fb_db = auth.firebaseDB()
+    fb_db.reference("v2/projects/{0}".format(project_id)).set(project_data)
+    fb_db.reference("projects/{0}".format(project_id)).set({})
```
Matthias-Schaub: If previous statement errors this statement could still be executed and delete all old projects and therefor we loos all old projects.
Either we wrap this in a transaction or delete old projects when we are sure the transfer was successful.

mapswipe_workers/scripts/archive_old_projects.py
```diff
@@ -0,0 +1,69 @@
...
+    First get all old projects.
+    Move project data to v2/projects in Firebase and
+    set status=archived.
+    Then delete all groups for a project.
+    Finally, delete old results, imports and announcements.
+    We don't touch the old user data in this workflow.
+    """
+
+    projects = get_old_projects()
+    for project_id in projects.keys():
```
laurentS: Agree with @Matthias-Schaub's comment. If possible, I would use a db transaction to enclose the 3 lines below, so the db will ensure everything works, or it will rollback the individual project. I guess you need to do something if the transaction fails, and that could be as simple as just logging the project id to whatever is easiest. This is a script you'll run once in your life, so don't overcomplicate it, you could just run it from one of the servers over ssh in a terminal and log with `console.log()`, just make sure it doesn't lose your data :)

mapswipe_workers/scripts/archive_old_projects.py
Hagellach37: I used a firebase transaction function

mapswipe_workers/scripts/archive_old_projects.py
Hagellach37: I used the firebase transaction function

PR #284, opened 2020-02-03
Configuration

Matthias-Schaub:
- Change paths handling.
    - Directory creation for api data is now been done by python-mapswipe_workers in
Python.
    - Create user Mapswipe in Dockerfile and use XDG Base Directory Specification for config and data directories. This will make it easier to deploy and run mapswipe_workers without using docker (e.g. locally ona computer during development or running tests).
- Create dirs for api data if not exists on each run of mapswipe_workers cli.
- If no Sentry configuration is provided continue without Sentry integration.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -55,15 +55,8 @@ services:
             - postgres
         command: mapswipe_workers --verbose run --schedule
         volumes:
-            - ./api-data:/var/lib/mapswipe_workers/api-data/:rw
-            - ./api-data/agg_results:/var/lib/mapswipe_workers/api-data/agg_results:rw
-            - ./api-data/groups:/var/lib/mapswipe_workers/api-data/groups:rw
-            - ./api-data/history:/var/lib/mapswipe_workers/api-data/history:rw
-            - ./api-data/projects:/var/lib/mapswipe_workers/api-data/projects:rw
-            - ./api-data/results:/var/lib/mapswipe_workers/api-data/results:rw
-            - ./api-data/tasks:/var/lib/mapswipe_workers/api-data/tasks:rw
-            - ./api-data/yes_maybe:/var/lib/mapswipe_workers/api-data/yes_maybe:rw
-            - ./api-data/hot_tm:/var/lib/mapswipe_workers/api-data/hot_tm:rw
+            - ./log:/var/lib/mapswipe_workers/:rw
+            - ./api-data:/home/mapswipe/.local/share/api-data/:rw
         restart: "no"
         networks:
             - mapswipe_workers
--- a/mapswipe_workers/Dockerfile
+++ b/mapswipe_workers/Dockerfile
@@ -2,6 +2,9 @@
 # Image includes python3.6, gdal-python, gdal-bin
 FROM osgeo/gdal:ubuntu-small-latest
 
+ENV LC_ALL=C.UTF-8
+ENV LANG=C.UTF-8
+
 # Install pip
 RUN apt-get update
 RUN apt-get --yes install python3-pip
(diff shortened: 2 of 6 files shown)


Hagellach37:
Hey @Matthias-Schaub,
I'm having problems checking this pull request. I am bit lost, where to start... Is there an established way how we would validate this, without merging into dev and then trying it on the server and fixing potential issues afterwards?

However, when running `docker-compose build mapswipe_workers` locally I get the following error:
```
Step 9/20 : RUN chown mapswipe_workers: /var/log/mapswipe_workers/
 ---> Running in 4b8c291f1d7c
chown: invalid spec: ‘mapswipe_workers:’
ERROR: Service 'mapswipe_workers' failed to build: The command '/bin/sh -c chown mapswipe_workers: /var/log/mapswipe_workers/' returned a non-zero code: 1
```

mapswipe_workers/Dockerfile
```diff
@@ -2,41 +2,36 @@
...
-RUN mkdir -p $data_dir"api-data/tasks/"
-RUN mkdir -p $data_dir"api-data/yes_maybe/"
-RUN mkdir -p $data_dir"api-data/hot_tm/"
+# Create user mapswipe
+RUN groupadd --system mapswipe
+RUN useradd --create-home --no-log-init --system --gid mapswipe mapswipe
+
+# Create directory for logs and make it accsesably for the mapswipe user
+RUN mkdir --parents /var/log/mapswipe_workers
+RUN chown mapswipe_workers: /var/log/mapswipe_workers/
```
Hagellach37: I think this line should be:
`RUN chown mapswipe:mapswipe /var/log/mapswipe_workers/`

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -41,6 +42,7 @@ def type_cast_value(self, ctx, value):
 @click.version_option()
 def cli(verbose):
     """Enable logging."""
+    create_directories()
```
Hagellach37: Hey,
maybe you could briefly explain to me again, why you think that it's better to create the directories here and not during the initialization of the docker container.

docker-compose.yaml
```diff
@@ -55,15 +55,8 @@ services:
...
-            - ./api-data/agg_results:/var/lib/mapswipe_workers/api-data/agg_results:rw
-            - ./api-data/groups:/var/lib/mapswipe_workers/api-data/groups:rw
-            - ./api-data/history:/var/lib/mapswipe_workers/api-data/history:rw
-            - ./api-data/projects:/var/lib/mapswipe_workers/api-data/projects:rw
-            - ./api-data/results:/var/lib/mapswipe_workers/api-data/results:rw
-            - ./api-data/tasks:/var/lib/mapswipe_workers/api-data/tasks:rw
-            - ./api-data/yes_maybe:/var/lib/mapswipe_workers/api-data/yes_maybe:rw
-            - ./api-data/hot_tm:/var/lib/mapswipe_workers/api-data/hot_tm:rw
+            - ./log:/var/lib/mapswipe_workers/:rw
+            - ./api-data:/home/mapswipe/.local/share/api-data/:rw
```
Hagellach37: I know that the lines above look really ugly. I think that I was struggling setting up the shared volumes between the `mapswipe_workers` container and the `api` container. Did you already double check that the api serves the data as we expect it?

mapswipe_workers/Dockerfile
Matthias-Schaub: Yes your are right. Thanks for pointing out.

docker-compose.yaml
Matthias-Schaub: I will double check.

Matthias-Schaub: I just run it on the dev server. At least projects/ and stats.csv has been written into the api-data dir by mapswipe_workers and successfully read by the api according to the timestamp.

PR #285, opened 2020-02-03
Update database rules for the leaderboard v2

### LINKED ISSUE #282: Read access to v2/users in firebase
The leaderboard needs read access to Firebase users. This needs to be adjusted in the firebase rules.

ericboucher:
Fixes #282

### DIFF
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -55,6 +55,7 @@
         }
       },
       "users": {
+        ".read": true,
         "$uid": {
           ".read": "auth != null && auth.uid == $uid",
           ".write": "auth != null && auth.uid == $uid"
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -57,7 +57,6 @@
       "users": {
         ".read": true,
         "$uid": {
-          ".read": "auth != null && auth.uid == $uid",
           ".write": "auth != null && auth.uid == $uid"
         },
         ".indexOn": [


firebase/database.rules.json
```diff
@@ -55,6 +55,7 @@
         }
       },
       "users": {
+        ".read": true,
         "$uid": {
           ".read": "auth != null && auth.uid == $uid",
```
Hagellach37: we might need to adjust (or remove) this line as well, since it will overwrite the other rule two lines above. Otherwise the leaderboard would only be able to get the user ids and not the actual data

Hagellach37: double checked it. The upper statement overwrites the lower one. I will remove it, since it's a bit confusing to have it when it has not effect.

ericboucher: I think it's fine but let me give it a shot

Hagellach37: sorry, my sentence was confusing.

Your statement is correct, but I will remove line 60. :)

Hagellach37:
ah and one more thing:
I overlooked that this merge was into `master`....

We set it up this way that we first merge into `dev` branch and then into `master`. This way we will be able to test things in dev first.

firebase/database.rules.json
ericboucher: Maybe remove it for the v1 data as well? It has a similar structure at the moment @Hagellach37

ericboucher:
@Hagellach37 noted for next time! Would you like me to also create a PR for `dev`?

Hagellach37:
no problem, i did the PR into dev already. :)

PR #288, opened 2020-02-03
Update db rule

Hagellach37:


mapswipe_workers/mapswipe_workers/base/base_project.py
```diff
@@ -217,11 +216,10 @@ def save_to_postgres(self, project, groups, groupsOfTasks):
 
         query_insert_project = '''
             INSERT INTO projects
-            VALUES (%s,%s,%s,ST_Force2D(ST_Multi(ST_GeomFromText(%s, 4326))),%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s);
+            VALUES (%s,%s,ST_Force2D(ST_Multi(ST_GeomFromText(%s, 4326))),%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s,%s);
```
ericboucher: This seems like an unrelated change @Hagellach37 ?

Hagellach37: hey,
this was a hotfix for the production. Due to this we could not create projects and to quickly fix it we addressed in the master branch. I am taking another look to validate that things are working as expected and will then release the new master to the back end. :)

PR #308, opened 2020-03-04
Travis

Matthias-Schaub:
- Extent travis configuration
- Rewrite configuration handling to using environment vars instead of json file.


Hagellach37:
`vitualenv` option only works with python 3.5

Hagellach37:
First time that a build passed: https://travis-ci.org/mapswipe/python-mapswipe-workers/builds/660864599?utm_medium=notification

Yeah!

Hagellach37:
@ericboucher would you be interested in doing a rough review. I recall that you once urged us to set up travis and also started yourself. :)

We got quite far now. The build is actually passing and we can add tests from now on.

mapswipe_workers/mapswipe_workers/definitions.py
```diff
@@ -1,34 +1,58 @@
...
-
-
-CONFIG_DIR = os.path.join(XDG_CONFIG_HOME, "mapswipe_workers")
-
-CONFIG_PATH = os.path.join(CONFIG_DIR, "configuration.json")
-
-CONFIG = load_config(CONFIG_PATH)
-
-SERVICE_ACCOUNT_KEY_PATH = os.path.join(CONFIG_DIR, "serviceAccountKey.json")
+IMAGE_URL = {
```
Hagellach37, line 15: just to double check:
is this getting us the right strings we need later for project creation?

docs/source/dev_setup.md
```diff
@@ -21,55 +31,65 @@ git checkout dev
...
-3. Install MapSwipe Workers using pip.
-4. Run it.
+Setup a local Postgres instance for MapSwipe Workers using the Dockerfile provided for development purposes (`postgres/Dockerfile-dev`). Make sure that the specified port is not in use already. If so, adjust the port (also in the `.env` file).
 
 ```bash
-python -m venv --system-site-packages venv
-source venv/bin/activate
-pip install --editable .
-mapswipe_workers --help
+docker build -t mapswipe_postgres -f ./postgres/Dockerfile-dev ./postgres
```
Hagellach37, line 63: any reasons why you are not using this instead?
```
docker-compose build postgres
docker-compose up -d postgres
```

mapswipe_workers/mapswipe_workers/utils/user_management.py
```diff
@@ -1,167 +1,157 @@
...
 
 def get_firebase_db(path, custom_arguments=None, token=None):
-    databaseName = CONFIG['firebase']['database_name']
-    database_url = f'https://{databaseName}.firebaseio.com'
-    request_ref = '{0}{1}.json?{3}auth={2}'.format(database_url, path, token, custom_arguments)
-    headers = {"content-type": "application/json; charset=UTF-8"}
-    request_object = requests.get(
-        request_ref,
-        headers=headers
+    databaseName = FIREBASE_DB
```
Hagellach37, line 101: maybe picky, but why `databaseName` instead of `database_name`

mapswipe_workers/mapswipe_workers/definitions.py
Hagellach37: in auth.py we use the `get_tileserver_function`

```
IMAGE_URL['bing']
('https://ecn.t0.tiles.virtualearth.net', '/tiles/a{quad_key}.jpeg?g=7505&mkt=en-US&token={key}')
```

mapswipe_workers/mapswipe_workers/auth.py
```diff
@@ -1,22 +1,30 @@
...
+)
+from mapswipe_workers.definitions import IMAGE_URL
 
 
 def get_api_key(tileserver: str) -> str:
     if tileserver == "custom":
         return None
     else:
-        return CONFIG["imagery"][tileserver]["api_key"]
+        return IMAGE_URL[tileserver]
```
Hagellach37, line 20: this should return the api key instead, but it return the url.
we need to load the api key for each tile server from `mapswipe_workers.config` above.

mapswipe_workers/mapswipe_workers/config.py
```diff
@@ -0,0 +1,26 @@
...
+
+IMAGE_BING_API_KEY = os.getenv("IMAGE_BING_API_KEY")
+IMAGE_DIGITAL_GLOBE_API_KEY = os.getenv("IMAGE_DIGITAL_GLOBE_API_KEY")
+IMAGE_ESRI = os.getenv("IMAGE_ESRI")
+IMAGE_ESRI_BETA = os.getenv("IMAGE_ESRI_BETA")
+IMAGE_MAPBOX_API_KEY = os.getenv("IMAGE_MAPBOX_API_KEY")
+IMAGE_MAXAR_PREMIUM_API_KEY = os.getenv("IMAGE_MAXAR_PREMIUM_API_KEY")
+IMAGE_MAXAR_STANDARD_API_KEY = os.getenv("IMAGE_MAXAR_STANDARD_API_KEY")
+
+SLACK_CHANNEL = os.getenv("SLACK_CHANNEL", default=None)
```
Hagellach37, line 24: it seems that the default value is always `None` if the variable doesn't exist.

Hagellach37:
Hey @Matthias-Schaub ,
I just went through the code. Most of it looks fine, I added some minor comments. But all in all ready to merge to dev then. :)

docs/source/dev_setup.md
Matthias-Schaub: Because for the dev setup there exists a different postgres Dockerfile (`/postgres/Dockerfile-dev `) without the backup setup. This Dockerfile is not included in the docker-compose yet. We could add it, but its only for development purposes.

mapswipe_workers/mapswipe_workers/utils/user_management.py
Matthias-Schaub: No reason. I changed it.

mapswipe_workers/mapswipe_workers/auth.py
Matthias-Schaub: Good catch. I created an IMAGE_API_KEYS dict with same structure to IMAGE_URLS to be imported from config.

mapswipe_workers/mapswipe_workers/definitions.py
Matthias-Schaub: No, you are right. This will return a tuple instead of a string. Corrected `("...", "...")` to `("..."+"...")`. Thanks.

mapswipe_workers/mapswipe_workers/config.py
Matthias-Schaub: True. Removed default argument.

PR #311, opened 2020-03-09
Fix archive projects

### LINKED ISSUE #310: Archive projects: max write limit
if you want to delete large parts in firebase (e.g. tasks for a project) the following error can occur:

```
firebase_admin.exceptions.InvalidArgumentError: Data to write exceeds the maximum size that can be modified with a single request.
```

We need to make sure that we delete step by step

This error occurred for project `-LvH36WJmWk7IKRB1zjH` in dev.

Hagellach37:
fix #310 bug.

### DIFF
--- a/mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
+++ b/mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
@@ -21,24 +21,28 @@ def archive_project(project_ids: list) -> None:
         fb_db.reference("v2/results/{0}".format(project_id)).set({})
 
         # get group keys for this project to estimate size in firebase
-        group_keys = list(
-            fb_db.reference("v2/tasks/{0}/".format(project_id)).get(shallow=True).keys()
-        )
-        chunk_size = 250
-        chunks = int(len(group_keys) / chunk_size) + 1
+        groups = fb_db.reference("v2/groups/{0}/".format(project_id)).get(shallow=True)
 
-        # delete groups and tasks in firebase for each chunk using the update function
-        for i in range(1, chunks):
-            logger.info(
-                "Delete max {1} groups and tasks of project with the id {0}".format(
-                    project_id, chunk_size
+        if not groups:
+            logger.info("no groups to delete in firebase")
+        else:
+            group_keys = list(groups.keys())
+            chunk_size = 250
+            chunks = int(len(group_keys) / chunk_size) + 1
+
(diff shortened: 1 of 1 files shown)


mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
```diff
@@ -15,24 +15,54 @@ def archive_project(project_ids: list) -> None:
     """
     for project_id in project_ids:
         logger.info("Archive project with the id {0}".format(project_id))
-        logger.info(
-            "Delete results, groups and tasks of project with the id {0}".format(
-                project_id
-            )
-        )
+        logger.info("Delete results of project with the id {0}".format(project_id))
```
laurentS: Why not use f-strings? If you're using python 3.6+ everywhere, they'd be easier to read. This line would be:
`logger.info(f"Delete results of project with the id {project_id}")`.
Have a look at https://realpython.com/python-f-strings/ for more explanations.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
```diff
@@ -15,24 +15,54 @@ def archive_project(project_ids: list) -> None:
...
+        )
         pg_db = auth.postgresDB()
-        sql_query = (
-            "UPDATE projects SET status = 'archived' "
-            + "WHERE project_id = '{0}'".format(project_id)
+        sql_query = """
+            UPDATE projects SET status = 'archived'
+            WHERE project_id = '{0}';
+        """.format(
+            project_id
```
laurentS: Are you sure that `project_id` is an integer? Is it possible that anything else can be injected?
Also, you should use f-strings here too, it's much easier to read :)

mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
Hagellach37: yep, we have been using this before as well. Don't know we didn't do it here already. Adjusted

mapswipe_workers/mapswipe_workers/firebase_to_postgres/archive_project.py
Hagellach37: we use save string formatting now and pass named parameters to the query instead of using string formatting

Hagellach37:
Thanks for the comments @laurentS

PR #315, opened 2020-03-10
bring master and dev to same level

Hagellach37:


Hagellach37:
basically nothing changed...

Issue #294, opened 2020-02-05
Add functionality: Delete Projects


Hagellach37:
From time to time projects get created but never worked on. We need the option to delete these projects. This can be done similar to archiving a project.

https://github.com/mapswipe/python-mapswipe-workers/blob/82f7d68cccd0759f28fdda34e546bccf9fc606dd/mapswipe_workers/mapswipe_workers/mapswipe_workers.py#L183-L203

Hagellach37:
Hey @Matthias-Schaub ,
sorry for being slow. but maybe we need to discuss this again. We figured out that all groups have been deleted (probably by accident) on the dev-firebase instance.

Let's double check that this can't happen whenever one uses this function with the wrong parameters. :)

Matthias-Schaub:
Hi, no worries. 

You are right. Groups in dev-firebase where deleted during testing of delete_projects module.

Currently tests for the right parameters (not None) are done in the cli (mapswipe_workers.py). During testing only the delete_projects module was tested without the checks of the cli. Thats the reason for the accidental deletion of all groups. Sorry for that.

I will add additional checks in the delete_projects module that the produced reference is valid, meaning it will nevere delete all projects, groups, tasks or results.

Matthias-Schaub:
Following expression will make sure that the reference to delete is at least 3 level deep, so that not all projects, groups, results or taks can be deleted:

`if not re.match("/v2/\w+/[a-zA-Z0-9|-|_]+", ref.path):`

PR #320, opened 2020-03-23
Delete project

Matthias-Schaub:
Implements issue #294

### DIFF
--- a/mapswipe_workers/mapswipe_workers/firebase_to_postgres/delete_project.py
+++ b/mapswipe_workers/mapswipe_workers/firebase_to_postgres/delete_project.py
@@ -59,3 +59,5 @@ def delete_project(project_ids: list) -> None:
         pg_db.query(sql_query, project_id)
         sql_query = "DELETE FROM projects WHERE project_id = {};".format(project_id)
         pg_db.query(sql_query, project_id)
+
+    return True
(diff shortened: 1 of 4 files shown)


mapswipe_workers/mapswipe_workers/firebase_to_postgres/delete_project.py
```diff
@@ -0,0 +1,34 @@
...
+    Deletes project, groups, tasks and results from Firebase and Postgres
+    """
+    for project_id in project_ids:
+        logger.info(
+            f"Delete project, groups, tasks and results of project: {project_id}"
+        )
+
+        fb_db = auth.firebaseDB()
+        fb_db.reference(f"v2/results/{project_id}").delete()
+        fb_db.reference(f"v2/tasks/{project_id}").delete()
```
Hagellach37: similar to the approach in archive projects we might need to think about splitting this up into chunks to avoid write size limit errors

Matthias-Schaub: Does this limit also apply to the delete function? I will try to find this out.

Matthias-Schaub: The limit also apllies to the delete functiion. I could test that. A soloution using chunks of data is now implemented (see also #310).

PR #324, opened 2020-03-31
Check if Firebase reference path is a valid one before deletion.

Matthias-Schaub:


Hagellach37:
Hey @Matthias-Schaub,
the firebase ref checking looks good. Could you take a look at the postgres string formatting part? And then we are ready to merge. :)

Hagellach37:
Hey @Matthias-Schaub,
I slightly changed the regex to include also projects starting with '-'. Other than that, things work. yeah!
I will merge this now, thanks for your work. :)

Issue #333, opened 2020-04-28
Simplify Tasking Manager Geometries


Hagellach37:
Simpler geometries from MapSwipe would better integrate with HOT's Tasking Manager and could help to reduce traffic, inprove user experience, and and reduce load for Tasking Manager api.

Adjust our workflow so that we:
* use bounding box instead of convex hull
* use only 7 digits for coordinates
* provide addtional output for project extent as geojson (the extent the project manager has used to create the project in the first place)

could look like this in the end:

![Screenshot from 2020-04-28 12-56-17](https://user-images.githubusercontent.com/7045979/80480439-f213e300-8950-11ea-8815-4ed69512c013.png)

We might be even able to set up a project in the Tasking Manager with a simple aoi and arbitrary tasks utilizing the Tasking Manager api. :)

xamanu:
Thanks, @Hagellach37 for checking on this. We just limited the standard size for file upload in the Tasking Manager to 1 MB (https://github.com/hotosm/tasking-manager/issues/2808). Please let us know if you suggest a different maximum upload size.

PR #336, opened 2020-04-28
Tasking manager geometries

Hagellach37:
This implements #333

* tasking manager geometries are simpler (bounding box instead of convex hull)
* use a coordinate precision of 7 digits
* we additionally export the project extent as geojson file during project creation

Further it fixes a small bug in setting up the logger.

@Matthias-Schaub could you review this in the coming days?

### DIFF
--- a/mapswipe_workers/scripts/add_project_geometries_to_api.py
+++ b/mapswipe_workers/scripts/add_project_geometries_to_api.py
@@ -19,7 +19,7 @@ def add_project_geometries_to_api():
     print(len(data))
 
     # save as geojson one by one
-    for i, project in enumerate(data):
+    for project in data:
         project_id = project[0]
         wkt_geom = project[1]
 
@@ -29,7 +29,7 @@ def add_project_geometries_to_api():
         try:
             geometries = [ogr.CreateGeometryFromWkt(wkt_geom)]
             geojson_functions.create_geojson_file(geometries, outfile)
-        except Exception as e:
+        except Exception:
             print(f"got an error for {project_id}")
             # just ignore if this fails
             pass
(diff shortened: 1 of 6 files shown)


Hagellach37:
Hey @Matthias-Schaub,
can you take a quick look at this one in the coming days? :)

mapswipe_workers/scripts/add_project_geometries_to_api.py
```diff
@@ -0,0 +1,38 @@
...
+        SELECT
+            project_id
+            ,ST_AsText(geom) as geom
+        FROM projects
+        """
+    data = pg_db.retr_query(sql_query)
+    print(len(data))
+
+    # save as geojson one by one
+    for i, project in enumerate(data):
```
Matthias-Schaub, line 22: Variable i is never used. Maybe `for project in data:` is enough?

mapswipe_workers/scripts/add_project_geometries_to_api.py
```diff
@@ -0,0 +1,38 @@
...
+        project_id = project[0]
+        wkt_geom = project[1]
+
+        outfile = (
+            f"{DATA_PATH}/api/project_geometries/project_geom_{project_id}.geojson"
+        )
+        try:
+            geometries = [ogr.CreateGeometryFromWkt(wkt_geom)]
+            geojson_functions.create_geojson_file(geometries, outfile)
+        except Exception as e:
```
Matthias-Schaub, line 32: `e` is never used: `except Exception` is enough.

pass is not needed in the exception. Only if nothing (no printing for example)/ no statement is in the exception clause then pass can be used.

Issue #330, opened 2020-04-19
Send a notification to Slack when a project reaches 100% progress


laurentS:
It should be fairly easy to send a notification from the firebase `incProjectProgress` function when progress reaches a given threshold:
- 90% progress: notify project manager channel that a project is almost complete, make sure there are more in the pipeline
- 100%: notify project managers they can replace the project with a new one

The only trick is to prevent sending messages more than once, so some sort of flag needs to be stored somewhere. It could simply be in the project object in firebase, like `notification_90_sent: true` once it's sent, and nothing before (so no need to change the project creation).

The code could either run in the progress update function, or as a "cron" job, checking progress of all projects every N minutes.

Hagellach37:
The back end workers are already calculating the progress for all active projects around every 15 minutes. So we have the information needed also in the back end and not only in firebase.

I like the idea of having the flag in firebase as an attribute of the project. We might not even need to set it up on project creation, since we can just assume that we didn't reach 90% or 100% when the attribute is missing.

PR #338, opened 2020-04-29
send project progress notifications #330

Hagellach37:
Hey @laurentS,
this one is for you. Will probably save you a lot of time in the future. :) You want to take a quick look?

### DIFF
--- a/mapswipe_workers/mapswipe_workers/mapswipe_workers.py
+++ b/mapswipe_workers/mapswipe_workers/mapswipe_workers.py
@@ -104,7 +104,8 @@ def run_firebase_to_postgres() -> list:
     update_data.update_user_data()
     update_data.update_project_data()
     project_ids = transfer_results.transfer_results()
-    [send_progress_notification(project_id) for project_id in project_ids]
+    for project_id in project_ids:
+        send_progress_notification(project_id)
     return project_ids
 
 
--- a/mapswipe_workers/mapswipe_workers/mapswipe_workers.py
+++ b/mapswipe_workers/mapswipe_workers/mapswipe_workers.py
@@ -27,7 +27,10 @@ from mapswipe_workers.project_types.change_detection.change_detection_project im
 from mapswipe_workers.project_types.footprint.footprint_project import FootprintProject
 from mapswipe_workers.utils import user_management
 from mapswipe_workers.utils.create_directories import create_directories
-from mapswipe_workers.utils.slack_helper import send_slack_message
+from mapswipe_workers.utils.slack_helper import (
+    send_slack_message,
+    send_progress_notification,
+)
 
 
 class PythonLiteralOption(click.Option):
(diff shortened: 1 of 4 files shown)


mapswipe_workers/mapswipe_workers/utils/slack_helper.py
```diff
@@ -27,14 +23,58 @@ def send_slack_message(message_type: str, project_name: str, project_id: str = N
             + "Make sure to activate the project using the manager dashboard.\n"
             + "Happy Swiping. :)"
         )
+        slack_client.chat_postMessage(channel=SLACK_CHANNEL, text=message)
     elif message_type == "fail":
```
laurentS, line 27: You could use an [enum](https://docs.python.org/3/library/enum.html) for the `message_type`, it's a bit more structured than using `str`, and you get better type checking and syntax completion, which in turn prevents typos, etc...
```python
from enum import Enum
class MessageType(Enum):
    FAIL = 1
    NOTIFICATION_90 = 2
    ...

Then you can check with `if message_type == MessageType.FAIL`, etc...
You'll need to import the class everywhere you use it, but it's probably a good way to define these centrally.

**Comment by Hagellach37** [comment]
@laurentS thanks for the comments. I updated my code. :)

File: mapswipe_workers/mapswipe_workers/utils/slack_helper.py [hunk 2]
```diff
@@ -57,24 +57,26 @@ def send_progress_notification(project_id: int):
...
-    logger.info(
-        f"{project_id} - progress: {progress},"
-        f"notifications: {notification_90_sent} {notification_100_sent}"
-    )
 
-    if progress >= 90 and not notification_90_sent:
-        # send notification and set value in firebase
-        fb_db.reference(f"v2/projects/{project_id}/notification_90_sent").set(True)
-        send_slack_message("notification_90", project_name, project_id)
+    if progress >= 90:
```
laurentS, line 64: I missed this the first time, but what data type is `fb_db.....get()` returning? I imagine a string? python might be unhappy about the comparison.

Matthias-Schaub: .get() usually returns a list or ordered dict.

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -101,6 +104,7 @@ def run_firebase_to_postgres() -> list:
     update_data.update_user_data()
     update_data.update_project_data()
     project_ids = transfer_results.transfer_results()
+    [send_progress_notification(project_id) for project_id in project_ids]
```
Matthias-Schaub, line 107: For me a simple for loop would be more readbale. Its unsual to do a list comprehension without saving the resulting list. At first it seems more like a potential bug.

mapswipe_workers/mapswipe_workers/utils/slack_helper.py
Hagellach37: in this case it returns an integer in this case, if if is a number in firebase.

PR #340, opened 2020-04-30
Dev

Matthias-Schaub:


.travis.yml
```diff
@@ -0,0 +1,71 @@
+# This is a special configuration file to run tests on Travis-CI via
```
Hagellach37, line 1: just to double check, did we set up travis also for the master branch?

docs/source/configuration.md
```diff
@@ -37,31 +38,47 @@ In addition to get access to Firebase a Service Account Key is required.
...
 
-On details of how the back-up works please refer to [Postgres Backup](backup.md).
+On details of how the back-up works please refer to [Postgres Backup](backup.html).
 
 Required environment variables are:
 - WALG_GS_PREFIX
 
-To gain access to Google Cloud Storage another Service Account Key is needed. Again refer to [Postgres Backup](backup.md) on how to create this file.
+To gain access to Google Cloud Storage another Service Account Key is needed. Again refer to [Postgres Backup](backup.html) on how to create this file.
 The Service Account Key (`serviceAccountKey.json`) should be saved to `postgres/serviceAccountKey.json`
```
Hagellach37, line 78: Is this still the case when we use the environment variables?

docs/source/dev_setup.md
```diff
@@ -21,60 +31,70 @@ git checkout dev
...
 
 
-### Install MapSwipe Workers Python Package
+### Postgres
 
-1. Create a Python virtual environment with `system-site-packages` option enabled to get access to GDAL/OGR Python packages
-2. Activate the virtual environment.
-3. Install MapSwipe Workers using pip.
-4. Run it.
+Setup a local Postgres instance for MapSwipe Workers using the Dockerfile provided for development purposes (`postgres/Dockerfile-dev`). Make sure that the specified port is not in use already. If so, adjust the port (also in the `.env` file).
```
Hagellach37, line 60: Maybe we can highlight here that the postgres image for development is different to the one used in production (or on the server) and that it doesn't write back ups. This was confusing for me the last time.

.travis.yml
Matthias-Schaub: Travis will build for all branches in which a .travis file exists.

docs/source/configuration.md
Matthias-Schaub: It is. Because we do not store the whole serviceAccountKey in an env. We store only the path to the file itself in the `GoogleAccountCredentials` variable. When using Dockerfiles each Dockerfile (mapswipe-workers and postgres/postgres_backup) will look localy (Same dir as the Dockerfile) for serviceAccountKey.json to copy it into the image. `GoogleAccountCredentials` has a default definied in docker-compose.yml

docs/source/dev_setup.md
Matthias-Schaub: Yes good hint. I will complement the docs on this.

.travis.yml
Hagellach37: true, but it need environment variables set up and did we do this?

docs/source/configuration.md
Hagellach37: mhh, okay but it is still the same file / the same key.

.travis.yml
Matthias-Schaub: Yes. All branches will use the same travis secrets/ environment variables.

docs/source/configuration.md
Matthias-Schaub: As before? Yes, still the same serviceAccountKey.json file/key. One from firebase for workers, one from google cloud storage for postgres.

PR #365, opened 2020-05-26
add usernames in firebase using python script

Hagellach37:


mapswipe_workers/python_scripts/add_usernames_in_firebase.py
```diff
@@ -0,0 +1,35 @@
...
+        user = auth.get_user(uid)
+        username = user.display_name
+        # only set username for users with display_name
+        if not username:
+            logger.info(f"user {uid} has no display_name in firebase.")
+        else:
+            ref = fb_db.reference(f"v2/users/{user.uid}/username")
+            ref.set(username)
+            logger.info(f"updated username for user {uid}: {username}")
+    except Exception:
```
laurentS, line 19: If you can find out what exception you'd get if the `uid` is not found, I'd do `except <thatException` so that in case something different/unexpected happens, you see it clearly

mapswipe_workers/python_scripts/add_usernames_in_firebase.py
```diff
@@ -0,0 +1,35 @@
...
+from mapswipe_workers.auth import firebaseDB
+from mapswipe_workers.definitions import logger
+
+
+def update_username(uid):
+    """Set username in Firebase DB based on auth.display_name for user id."""
+    fb_db = firebaseDB()
+    try:
+        user = auth.get_user(uid)
+        username = user.display_name
```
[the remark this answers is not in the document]
Hagellach37: correct, this parameter is `display_name` in python.

mapswipe_workers/python_scripts/add_usernames_in_firebase.py
Hagellach37: added correct exception

Issue #371, opened 2020-06-08
reduce download from firebase for back end workers


Hagellach37:
make sure that the back end only downloads data for some users and not all information, e.g. NOT all contributions for all users. This will dramatically reduce downloads from Firebase.

test queries for reference:

this query might not work correctly and unexpectedly downloads all users...
https://msf-mapswipe.firebaseio.com/v2/users.json?orderBy=%22created%22&StartAt=2020-06-08T15:40:13.653Z

Another complementary improvement would be to move the detailed user contribution stats to another endpoint in Firebase, e.g. `v2/userContributions/` This will also reduce the downloads from the leaderboard.

* [x] adjust `update_user_data` function
* [x] adjust firebase functions to add contributions to new endpoint
* [x] make sure that all contributions stats are correctly transfered in firebase

Hagellach37:
the most dramatic download issue had been fixed already in https://github.com/mapswipe/python-mapswipe-workers/releases/tag/v2.2.1, but by "accident". 

Still it would be much better to move the user contribution data to another path in Firebase to make it more intuitive to avoid such downloads.

Hagellach37:
I will open a new issue for this and describe better...

Issue #360, opened 2020-05-21
Update Cloud Functions to Node.js 10


Hagellach37:
By February 15, 2021, all developers using the Node.js 8 runtime should update to the Node.js 10 runtime and update to the “Blaze” pay-as-you-go billing plan. Developers will no longer be able to create new functions or update existing functions that use Node.js 8.

By March 15, 2021, Node.js 8 runtime will be shut down for Cloud Functions for Firebase. Functions using Node.js 8 will be disabled after this date.

Ideally this would be done together with #131

Hagellach37:
this is also part of #401

PR #372, opened 2020-06-08
Fix update users download

Hagellach37:
This introduces a new endpoint in firebase for user contributions `v2/userContributions`. Before this user contributions have been stored here `v2/users/{user_id}/contributions`. This resulted in unnecessary data download by the back end workers and the leader board. #371 

This pull request updates the firebase functions to set the user contributions to `v2/userContributions` whenever new results are uploaded to Firebase. In the state here, the firebase functions still update also the old path `v2/users/{user_id}/contributions`. This can be changed once all MapSwipe clients can deal with it. (The client queries the user contributions path to select groups a user has not been worked on already.) 
Additionally, this pull request comes with a script to copy the "old" user contributions to the new path in Firebase.

Firebase Functions are updated to Node.js 10. #360 

There is a document how to deploy these changes here: `docs/source/user_contributions.md`

### DIFF
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -47,6 +47,9 @@ exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{use
     const groupContributionsRef         = admin.database().ref('/v2/userContributions/'+context.params.userId+'/'+context.params.projectId +'/'+context.params.groupId)
     const totalTimeSpentMappingRef      = admin.database().ref('/v2/users/'+context.params.userId+'/timeSpentMapping')
 
+    const contributionsRefOld           = admin.database().ref('/v2/users/'+context.params.userId+'/contributions/'+context.params.projectId)
+    const groupContributionsRefOld      = admin.database().ref('/v2/users/'+context.params.userId+'/contributions/'+context.params.projectId +'/'+context.params.groupId)
+
     const timestampRef          = admin.database().ref('/v2/results/'+context.params.projectId+'/'+context.params.groupId+'/'+context.params.userId+'/timestamp')
     const startTimeRef          = admin.database().ref('/v2/results/'+context.params.projectId+'/'+context.params.groupId+'/'+context.params.userId+'/startTime')
     const endTimeRef            = admin.database().ref('/v2/results/'+context.params.projectId+'/'+context.params.groupId+'/'+context.params.userId+'/endTime')
@@ -123,6 +126,23 @@ exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{use
             }
         })
     promises.push(contributions)
+
+    // this can removed once all clients don't need this data in Firebase
+    const contributionsOld = groupContributionsRefOld.once('value')
+        .then((dataSnapshot) => {
+            if (dataSnapshot.exists()) {
+                return null
+            }
+            else {
+            const data = {
+                'timestamp': result['timestamp'],
+                'startTime': result['startTime'],
+                'endTime': result['endTime']
+             }
+             return groupContributionsRefOld.set(data)
+            }
+        })
+    promises.push(contributionsOld)
 
     // // TODO: Does not work
     // const timeSpentMapping = timeSpentMappingRef.set((timeSpentMapping) => {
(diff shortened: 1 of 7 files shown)


firebase/functions/package.json
```diff
@@ -19,7 +19,7 @@
     "eslint-utils": "^1.4.1"
   },
   "engines": {
-    "node": "8"
+    "node": "10"
```
laurentS, line 22: This is a pretty big change, with the potential to break a lot of code :)
Have you tested everything with node 10?

firebase/functions/index.js
```diff
@@ -124,6 +167,22 @@ exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{use
...
     promises.push(contributions)
 
+    const contributionsNew = groupContributionsRefNew.once('value')
+        .then((dataSnapshot) => {
+            if (dataSnapshot.exists()) {
+                return null
+            }
+            else {
+            const data = {
+                'timestamp': result['timestamp'],
```
laurentS, line 177: looking at the app code, it seems that `timestamp` and `endTime` are the same value, maybe we should look into throwing `timestamp` away... unless you're modifying one of them in the backend functions?

mapswipe_workers/python_scripts/copy_user_contributions_in_firebase.py
```diff
@@ -0,0 +1,46 @@
...
+    return uid_list
+
+
+def copy_user_contributions(uid):
+    """Copy user contributions in Firebase from `users` to `userContributions`."""
+    fb_db = firebaseDB()
+    try:
+        ref = fb_db.reference(f"v2/users/{uid}/contributions")
+        user_contributions = ref.get()
+        # only set username for users with display_name
```
laurentS, line 19: I don't understand this comment. Is it a leftover from a previous version?

mapswipe_workers/python_scripts/copy_user_contributions_in_firebase.py
```diff
@@ -0,0 +1,46 @@
...
+                    f"updated user contributions for user {uid}, project {project_id}"
+                )
+
+            # TODO: uncomment when we want to delete old contributions from firebase.
+            #  This should be done after all mapswipe clients can deal with it.
+            # ref = fb_db.reference(f"v2/users/{uid}/contributions/")
+            # ref.set(None)
+
+    except ValueError:
+        logger.info(f"could copy user contributions in firebase for {uid}.")
```
laurentS, line 36: I think `not` is missing in the error message :)

firebase/functions/package.json
Hagellach37: mhh good questions, however I'm not really sure how to check properly. The only thing I can do is to check it somehow manually for the dev version of the app and look at all the places in firebase which should be updated.

But, we have to update to Node.js 10 by February 15, 2021 (#360 ). Since firebase is not allowing "old" functions which are written in Node.js 8 from then on.

mapswipe_workers/python_scripts/copy_user_contributions_in_firebase.py
Hagellach37: this comment should be removed. correct.

firebase/functions/index.js
Hagellach37: This is also a good comment and could lead us to a general discussion here. Do we actually need this information in firebase for each user?

Initially, we wanted to have this information to quanitify how much time a users spend on a specific group. But maybe it would be easier to put this information into the results generated by the app. E.g. by adding an attribute "timeSpent" instead of having "timestamp" and "endTime" which are the same.

And then we might even think about why we store the timestamps in firebase at all and not only the ids of the projects and groups a user has worked on. Having only this would also reduce data a lot and we currently do not use this information in firebase at all...

mapswipe_workers/python_scripts/copy_user_contributions_in_firebase.py
Hagellach37: correct

Hagellach37:
Hey @laurentS,
thanks for your comments. It seems that we might need to have a more general discussion on this information.

* Which information is actually needed for what purpose?
* where is the best place to store it?

Hagellach37:
I need to make sure that the new firebase rules do not break old apps from transitioning to v2.

they need to get access to `users` or we do the transfer from the back end side

Hagellach37:
okay, let's forget about this pull request. After discussing with Laurent, I will open a new one and use a different approach...

Issue #368, opened 2020-06-02
Generate Stats: Memory Error in get_agg_results_by_user_id


Hagellach37:
The mapswipe workers generate statistics for the projects. One of the generated stats is about the users (`agg_results_by_user_id`). This happens here:

https://github.com/mapswipe/python-mapswipe-workers/blob/989eecb4aa501ffd7af47a0e01c0afe1cef4dcf8/mapswipe_workers/mapswipe_workers/generate_stats/user_stats.py#L30

When running the user stats for all projects, we get a memory error for the project with the id `2`. This might be related to the fact that for the very first MapSwipe projects (which have been set up in 2016), there was an issue with the correct task scheduling and there are several 1000 users per tasks.

The implementation of the stats generation uses the `pandas` library. To avoid the MemoryError and speed up the stats generation in general @laurentS had already some ideas:
* my guess is that you're mixing pandas native calls and python functions, so you have a lot of back and forth between these 2 spaces. ideally, if you could write the function in pure pandas, it'd run much faster, and with much less memory.
* for L67: `agg_results_by_user_id_df["simple_agreement_score"] = agg_resuts_by_user_id_df["agreeing_contributions"] / (agg_resuts_by_user_id_df["agreeing_contributions"] + agg_resuts_by_user_id_df["disagreeing_contributions"]`) should work, instead of doing apply, and it should be much faster
* could you add extra columns to your df, with intermediate results just using pandas native functions?
* https://stackoverflow.com/questions/49161120/pandas-python-set-value-of-one-column-based-on-value-in-another-column

Hagellach37:
comments from Laurent:
* my guess is that you're mixing pandas native calls and python functions, so you have a lot of back and forth between these 2 spaces. ideally, if you could write the function in pure pandas, it'd run much faster, and with much less memory.
* for L67: `agg_results_by_user_id_df["simple_agreement_score"] = agg_resuts_by_user_id_df["agreeing_contributions"] / (agg_resuts_by_user_id_df["agreeing_contributions"] + agg_resuts_by_user_id_df["disagreeing_contributions"]`) should work, instead of doing apply, and it should be much faster
* could you add extra columns to your df, with intermediate results just using pandas native functions?

Hagellach37:
getting another error now as well:
```
/usr/local/lib/python3.6/dist-packages/mapswipe_workers/generate_stats/user_stats.py:25: RuntimeWarning: invalid value encountered in long_scalars
[...]

PR #377, opened 2020-06-09
Fix memory error user stats

Hagellach37:
Implement #368


Hagellach37:
Hey @laurentS  and @Matthias-Schaub ,
I'm not sure who of you might be the best to look at this one. On the one hand Laurent gave some good comments on user stats last week or so. Matthias has started to implemented the unittests. 

I tried to combine both here. :) 

And please let me know, if you think I'm pushing to many pull requests for your review.

PR #378, opened 2020-06-11
Tests - Restructure fixtures

Matthias-Schaub:


Hagellach37:
ah maybe just a minor comment, we need to change the command in travis to run the tests correctly within the new structure

Hagellach37:
There is an issue here with the GDAL install in travis. I will merge this pull request now and we can address this issue in another one, see also #379

PR #380, opened 2020-06-15
Unittests

Matthias-Schaub:


.travis.yml
```diff
@@ -37,16 +37,17 @@ install:
   - mkdir --parents ~/.config/mapswipe_workers
   - mkdir --parents ~/.local/share/mapswipe_workers
   - echo "$SERVICE_ACCOUNT_KEY" > ~/.config/mapswipe_workers/serviceAccountKey.json
-  - docker build --tag mapswipe_postgres --file postgres/Dockerfile-dev postgres/
+  - docker build --tag mapswipe_postgres --file postgres/Dockerfile-dev postgres/  # Dockerfile-dev is used because backup routine is not needed for travis builds
+  - docker-compose build mapswipe_workers
 
 before_script:
   - docker run --detach --publish 5432:5432 --name mapswipe_postgres -e POSTGRES_DB="$POSTGRES_DB" -e POSTGRES_USER="$POSTGRES_USER" -e POSTGRES_PASSWORD="$POSTGRES_PASSWORD" mapswipe_postgres
   - docker-compose up --build --detach firebase_deploy
```
Hagellach37, line 45: which firebase instance is used here again? `dev` or `test`. and do I still need to setup the `test` firebase correctly?

.travis.yml
```diff
@@ -37,16 +37,17 @@ install:
...
+  - docker run python-mapswipe-workers_mapswipe_workers python -m unittest -v tests.integration.test_gdal
```
Hagellach37, line 46: is there a reason why run this as the `before` script and not in the `script` part?

Matthias-Schaub: no particular reason. Remove `before_script` for the sake of simplicity.

.travis.yml
Matthias-Schaub: At the moment the test or cli instance of firebase is used.

Issue #382, opened 2020-06-22
Changes in Results Timestamp


Hagellach37:
The results in Firebase change slightly regarding the timestamps.

This needs to be accounted for by:
* Firebase Functions
* Firebase to Postgres in MapSwipe workers

laurentS:
Just repeating my comment in https://github.com/mapswipe/mapswipe/pull/237 for easier tracking.

The commits in the PR don't change anything that the app sends to the backend. It only changes their names internally, and removes the `timestamp` key that is sent along with `startTime` and `endTime` in the results. So there are no new attributes. However if the backend relies on `timestamp` then you should change that to use `endTime` instead, as their values are the same.

PR #383, opened 2020-06-22
Timestamp attribute

Hagellach37:
Adjust back end to account for slight changes in app. #382 

https://github.com/mapswipe/mapswipe/pull/237


Hagellach37:
the timestamp attribute is stored in postgres, I think it's safer for now to just leave the DB schema as it is.

Issue #385, opened 2020-06-22
Tutorial: Bug in create change detection tutorial


Hagellach37:
so I'm looking at the CD tutorial, and adding the missing bits of code so that it works again. There are some weird "bugs" in the tutorial data on firebase. For instance, the status was "tutorial" instead of the name of the tutorial like it is now on built area. Right now, I'm looking at the pagination code, which decides which page it is on, based on taskX . In built area, the taskX values are "continuous", so 100, 101, 102, 103, etc.. which allows to compare the calculated value based on progress through the screens with the values in firebase. On the CD tuto, the values are even values only, 100, 102, 104, 106, so that comparison does not work. Is this supposed to be like this, or did you change something in the backend code that was not reflected for CD?

PR #386, opened 2020-06-22
adjust tutorial for change detection and add tests

Hagellach37:
this implements #385 and accounts for the different screen for change detection project in the tutorial.


Hagellach37:
@laurentS this should fix the wrong `taskX` and `taskY` for change detection tutorials.

However, it seems that there is another bug. The imagery used for the tutorial is hosted on OpenAerialMap. I can't get the tiles to show up and it seems that the tiles are not available as expected.

we use these two images from OpenAerialMap:
* IOM_Rohingya_refugee_camp_Kutupalong_Area_20180424
* Kutupalong-Balukhali-Thangkhali-Hakimpara-Jamtoli-Moynarghona_(Potibonia)

(it's somehow not really possible to link directly to this in OAM...)

Hagellach37:
and this being said... I found my mistake. The zoom level was set to 18 instead of 18...
Images should be displayed now. :)

Hagellach37:
not relevant anymore

Issue #392, opened 2020-07-10
projects for teams


Hagellach37:
Make projects available for users which are part of a team. Users that are part of a team will see more projects in the app than "normal" users, if there are specific team projects set. Whereas it is most desirable that MapSwipe projects are open for everyone to map, some organisations have requested this option, e.g. to engage volunteers in bigger MapSwipeathons or for targeted and potentially paid mapping campaigns.

This requires several additions from the back end side:

### Firebase
* [x] adjust firebase database rules
    * users can read their team id, but can't overwrite it
    * only project managers can write team attributes in firebase
    * users can read `teamName`, but not other attributes such as `teamToken`
* [x] write tests for firebase rules 
* [x] add new data endpoint for teams at `v2/teams/teamId`
* data structure:  ```{
  "teamName": "Benni's Super Swipers",
  "teamToken": "ab234ef449ccb98e7a"
  }```

### User / Team Management
* [x] write function to set up a team using the mapswipe_workers cli team-management
    * input: team name
    * teamToken will be generated, e.g. such as a UUID using [this lib](https://github.com/uuidjs/uuid)
* [x] write a function to renew token of a team in mapswipe_workers cli team-management
* [x] write function to delete a team  in mapswipe_workers cli team-management
* [x] write function to add user to a team with given id using the mapswipe_workers cli
* [x] write function to remove all users to a give team using the mapswipe_workers cli
* [x] add basic tests:
    * create a team, then delete it. There should be nothing left
    * create a team, make sure the data is there, and the UUID has the format you expect
    * create a team, update the team token, make sure the new token is not the same as the original one

### Project Creation
* [x] adjust project creation in manager dashboard
    * add team id to project
* [x] check for teamId in projectDraft during project creation and set status accordingly
    * set status for private projects after creation to `private_inactive`

### Project Management
* [x] show all users of a team in manager dashboard
* [x] show that projects are private in manager dashboard project overview page
* [x] make sure that you can set a private project not to public by accident, valid status attributes for private projects  `private_active`, `private_finished`, `private_inactive`

laurentS:
Here are some test ideas to validate the permission, based on the queries the app actually uses:
```python
# all the following should probably be done 4 times:
# - as a regular, non-project-manager user without teamId
# - as a project-manager user without teamId
# - as a regulat, non-project-manager user with a teamId
# - as a project-manager user with a teamId

[...]

PR #393, opened 2020-07-11
Feature private projects

Hagellach37:
This sets the basic structure needed from MapSwipe back end side to allow for private projects. This refers to #392 

* adds functions to mapswipe_workers cli to set up new teams and manage which users are members
* extends manager dashboard to create and display private projects and to display teams including their members
* adjusts firebase database rules to account for new endpoint at `v2/teams`
* adjusts project creation workflow to set status of private projects to `private_inactive` once created

Team set up is currently handled only by mapswipe_workers cli. There you can also renew the `teamToken` attribute if needed.

### DIFF
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -56,7 +56,7 @@
           ".read": "auth.token.projectManager === true",
           "teamName": {
             // team members read teamName
-            ".read": "(root.child('/v2/users/'+auth.uid+'/teamId').val() === $teamId) || (auth.token.projectManager === true)"
+            ".read": "(root.child('/v2/users/'+auth.uid+'/teamId').val() === $teamId)"
           },
           ".indexOn": [
             "teamName"
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -5,17 +5,20 @@
     "v2": {
       "announcement": {
         ".write": false,
-        ".read": true
+        ".read": "auth != null"
       },
       "projects": {
-        ".write": false,
-        ".read": "auth != null",
+        // project managers can read/write all attributes
+        ".write": "auth.token.projectManager === true",
+        ".read": "auth.token.projectManager === true",
         "$project_id": {
-          "status": {
-            ".write": "auth.token.projectManager === true"
+          "teamId": {
+            // only team members can read
+            ".read": "root.child('/v2/users/'+auth.uid+'/teamId').val() === data.val()"
           },
-          "isFeatured": {
-            ".write": "auth.token.projectManager === true"
+          "$other": {
+            // all users can read, project managers can write
+            ".read": "auth != null"
           }
         },
         ".indexOn": [
@@ -23,6 +26,7 @@
           ]
       },
       "projectDrafts": {
+        // only project managers can write projectDrafts
         ".read": false,
         ".write": "auth.token.projectManager === true",
         ".indexOn": [
@@ -30,8 +34,9 @@
         ]
       },
       "groups": {
+        // users can read tasks, nobody can write
         ".write": false,
-        ".read": true,
+        ".read": "auth != null",
         "$project_id": {
           ".indexOn": [
             "finishedCount",
@@ -40,15 +45,23 @@
         }
       },
       "tasks": {
+        // users can read tasks, nobody can write
         ".write": false,
-        ".read": true
+        ".read": "auth != null"
       },
       "teams": {
-        ".write": "auth.token.projectManager === true",
-        ".read": "auth.token.projectManager === true",
-        ".indexOn": [
-          "teamName"
-        ]
+        "$teamId": {
+          // project managers can read/write all attributes
+          ".write": "auth.token.projectManager === true",
+          ".read": "auth.token.projectManager === true",
+          "teamName": {
+            // team members read teamName
+            ".read": "(root.child('/v2/users/'+auth.uid+'/teamId').val() === $teamId) || (auth.token.projectManager === true)"
+          },
+          ".indexOn": [
+            "teamName"
+          ]
+        }
       },
       "results": {
         ".write": false,
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -91,56 +91,12 @@
         ]
       }
     },
-    "groups": {
-      ".write": false,
-      ".read": true,
-      "$project_id": {
-        "$group_id": {
-          ".indexOn": [
-            "distributedCount",
-            "completedCount"
-          ],
-          ".write": "auth != null",
-          "completedCount": {
-            ".write": "auth != null"
-          }
-        },
-        ".indexOn": [
-          "distributedCount",
-          "completedCount"
-        ]
-      }
-    },
-    "imports": {
-      ".read": false,
-      ".write": true,
-      ".indexOn": [
-        "complete"
-      ]
-    },
-    "projects": {
-      ".write": false,
-      ".read": true
-    },
-    "announcement": {
-      ".write": true,
-      ".read": true
-    },
-    "results": {
-      ".write": false,
-      ".read": true,
-      "$task_id": {
-        "$user_id": {
-          ".write": "auth != null && auth.uid == $user_id"
-        }
-      }
-    },
+    // leaving this here, since version before v2 pull data from there
     "users": {
       "$uid": {
         ".read": "auth != null && auth.uid == $uid",
         ".write": "auth != null && auth.uid == $uid"
-      },
-      ".read": true
+      }
     }
   }
 }
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -81,7 +81,7 @@
           "teamId": {
             ".write": false
           },
-          "$other" {
+          "$other": {
             ".write": "auth != null && auth.uid == $uid"
           }
         },
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -78,26 +78,11 @@
         ".read": true,
         ".write": false,
         "$uid": {
-          "username": {
-            ".write": "auth != null && auth.uid == $uid"
-          },
-          "created": {
-            ".write": "auth != null && auth.uid == $uid"
-          },
-          "lastAppUse": {
-            ".write": "auth != null && auth.uid == $uid"
-          },
-          "projectContributionCount": {
-            ".write": "auth != null && auth.uid == $uid"
-          },
-          "groupContributionCount": {
-            ".write": "auth != null && auth.uid == $uid"
-          },
-          "taskContributionCount": {
-            ".write": "auth != null && auth.uid == $uid"
-          },
           "teamId": {
             ".write": false
+          },
+          "$other" {
+            ".write": "auth != null && auth.uid == $uid"
           }
         },
         ".indexOn": [
--- a/firebase/database.rules.json
+++ b/firebase/database.rules.json
@@ -43,6 +43,13 @@
         ".write": false,
         ".read": true
       },
+      "teams": {
+        ".write": "auth.token.projectManager === true",
+        ".read": "auth.token.projectManager === true",
+        ".indexOn": [
+          "teamName"
+        ]
+      },
       "results": {
         ".write": false,
         ".read": false,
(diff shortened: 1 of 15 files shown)


mapswipe_workers/mapswipe_workers/utils/user_management.py
```diff
@@ -158,3 +158,52 @@ def permission_denied(request_object):
...
+            user = auth.get_user_by_email(email)
+        except auth.UserNotFoundError:
+            raise CustomError(f"can't find user in firebase: {email}")
+
+        # get teamId from firebase
+        team_id = fb_db.reference(f"v2/users/{user.uid}/teamId").get()
+
+        # remove teamId attribute for user in firebase
+        ref = fb_db.reference(f"v2/users/{user.uid}")
+        ref.update({"teamId": None})
```
laurentS, line 204: Does this result in the `teamId` attribute having a null value, or is it removed entirely from the db?

mapswipe_workers/mapswipe_workers/utils/user_management.py
```diff
@@ -158,3 +158,52 @@ def permission_denied(request_object):
             return True
         else:
             raise HTTPError(e, request_object.text)
+
+
+def add_team(email, team_id):
```
laurentS, line 163: I find this function name a bit confusing, I expect it to create a new team, just from the name. Could you rename it to something like `add_user_to_team`? Same thought about the function `remove_team` below.

mapswipe_workers/mapswipe_workers/utils/user_management.py
Hagellach37: basically this is the same, we set a null value in firebase and this results in the attribute being removed. It's just another way of deleting an attribute in firebase, but with update you could theoretically update several attributes at once.

laurentS: Ok, makes sense. I was mostly asking to understand how to approach this on the app side. So I can mostly assume that if `teamId` is present, it has a valid value (ie, not `null`).

mapswipe_workers/mapswipe_workers/utils/user_management.py
Hagellach37: agreed

manager_dashboard/manager_dashboard/js/project-management.js
```diff
@@ -3,68 +3,159 @@ function addEventListeners(status) {
...
+                    btn1 = addButton(data.key, data.val().status, "inactive")
+                    btn2 = addButton(data.key, data.val().status, "finished")
+                    row_array.push(btn1.outerHTML + btn2.outerHTML)
+                    btn = document.createElement('button')
+                    btn.id = data.key
+                    btn.classList.add("btn")
+                    btn.classList.add("btn-warning")
+                    btn.classList.add("change-isFeatured")
+                    btn.classList.add("isFeatured-"+data.val().isFeatured)
+                    if (data.val().isFeatured === true) {
```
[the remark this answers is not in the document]
Hagellach37: true

manager_dashboard/manager_dashboard/js/project-management.js
```diff
@@ -3,68 +3,159 @@ function addEventListeners(status) {
...
+                      btn.innerHTML = 'set to "false"'
+                      row_val = "<b>"+data.val().isFeatured+"</b>"
+                      row_array.push(row_val + "<br>" + btn.outerHTML)
+                    } else if (data.val().isFeatured === false) {
+                      btn.innerHTML = 'set to "true"'
+                      row_val = data.val().isFeatured
+                      row_array.push(row_val + "<br>" + btn.outerHTML)
+                    }
+                    row_array.push(row_val + btn.outerHTML)
```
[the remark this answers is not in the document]
Hagellach37: I added a function to create the button to change the isFeatured attribute. This makes it easier to read. Still not perfect maybe, but okay for now?

manager_dashboard/manager_dashboard/js/project-management.js
```diff
@@ -104,20 +203,17 @@ function updateTableView() {
...
-    getProjects("active")
-    getProjects("inactive")
-    getProjects("finished")
-    getProjects("archived")
-
   console.log('updated table view')
 }
 
 
 function changeProjectStatus() {
```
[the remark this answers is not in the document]
Hagellach37: I added more detailed comments to the functions.

firebase/database.rules.json
```diff
@@ -43,6 +43,13 @@
         ".write": false,
         ".read": true
       },
+      "teams": {
+        ".write": "auth.token.projectManager === true",
+        ".read": "auth.token.projectManager === true",
```
Hagellach37, line 48: @laurentS this might be a discussion point for the app side as well: do we plan to show the name of a team to the users of the app? currently only project managers can get the name from firebase.

Hagellach37:
Hey @laurentS I adjusted the parts you mentioned above. Do you have already something in mind regarding "add a bit of testing" for the CLI functions? :)
It's probably best do address this immediately in this PR and not postpone it to any time too late. ;)

laurentS:
> Hey @laurentS I adjusted the parts you mentioned above. Do you have already something in mind regarding "add a bit of testing" for the CLI functions? :)
> It's probably best do address this immediately in this PR and not postpone it to any time too late. ;)

Good idea to do it now :) I think I would just do a few basic scenarii, to make sure you have what you expect:
- create a team, then delete it. There should be nothing left
- create a team, make sure the data is there, and the UUID has the format you expect
- create a team, update the team token, make sure the new token is not the same as the original one

I think something along these lines is enough to start with. And if there are regressions down the line, it's easier to just add a test for that specific case.

firebase/database.rules.json
laurentS: I think it would be good to show it, just as a verification that everything is working ok. I suspect some users might be anxious that they're correctly registered as part of the team, if their stats depend on it. We could just add a line in the "profile" screen saying `Your team: xxxxx`. It would probably save project managers a bunch of emails asking to confirm team memberships. Also, would project names include the team name in them? I suspect it might help if we can add tag somewhere on the project to say it is part of "team X".

Hagellach37: okay, sounds good. I can then adjust the rules accordingly.
For the projects we only add the `teamId` not the  `teamName`, but once you have access to the teams you could easily display it in the app as well at the project overview page. 
However, we don't want that users can access the `teamToken`, right?

laurentS: I can't think of a situation where a user who's already part of a team would need the token, since its purpose was for registration only. If you're able to restrict access to the token, great, but if it's too complicated with firebase rules, then I wouldn't worry too much about it. In theory, if a user is part of a team, it is because they already had the token in the first place :) (I'm oversimplifying, as that ignores token changes)

firebase/database.rules.json
```diff
@@ -5,33 +5,38 @@
...
-        ".read": "auth != null",
+        // project managers can read/write all attributes
+        ".write": "auth.token.projectManager === true",
+        ".read": "auth.token.projectManager === true",
         "$project_id": {
-          "status": {
-            ".write": "auth.token.projectManager === true"
+          "teamId": {
+            // only team members can read
+            ".read": "root.child('/v2/users/'+auth.uid+'/teamId').val() === data.val()"
```
Hagellach37, line 17: @laurentS this is the simple test to check if the given `teamId`of the project (`data.val()`) and the teamId of the user match.

firebase/database.rules.json
Hagellach37: with the new firebase db rules, users can only read the `teamName` attribute of the team they are a member of

mapswipe_workers/tests/integration/test_firebase_db_rules.py
```diff
@@ -0,0 +1,44 @@
...
+
+        # project manager and team member
+
+        pass
+
+    def tearDown(self):
+        # tear down up 4 users with specified roles
+        pass
+
+    def test_normal_user(self):
```
Hagellach37, line 26: @laurentS just a quick on the structure of the tests: not sure which way round is nicer? test based on user role or based on endpoint?

just seeing this now it seems that by endpoint might be easier to understand.

laurentS: Based on my comment [here](https://github.com/mapswipe/python-mapswipe-workers/issues/392#issuecomment-658192603) I would write a function that pings each endpoint in turn, with a user passed as argument, and call it for each of the users you're interested in. Actually, from my perspective, you don't even care about the results, you just want to make sure you're not getting errors, so you could probably have a function that tests for error, given an endpoint, and then use that in the first function I described above?

```python
def test_endpoint(user, endpoint):
   # run the query and check for error

def test_all_app_endpoints(user):
    test_endpoint(user, endpoint1)
    test_endpoint(user, endpoint2)
    # ...
```

mapswipe_workers/tests/integration/test_firebase_db_rules.py
```diff
@@ -0,0 +1,329 @@
...
+        user_management.set_project_manager_rights(email)
+
+    # set team member attribute
+    if team_member:
+        user_management.add_user_to_team(email, team_id)
+
+    return user
+
+
+def sign_in_with_email_and_password(email: str, password: str) -> dict:
```
laurentS, line 77: I think the return value should be an uppercase `Dict` instead of `dict`. I'm not sure what happens with the lowercase version, I can't find any docs about it. The [mypy official cheatseet](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#built-in-types) uses the uppercase version though.

mapswipe_workers/tests/integration/test_firebase_db_rules.py
```diff
@@ -0,0 +1,329 @@
...
+    api_key = FIREBASE_API_KEY
+    request_ref = (
+        "https://identitytoolkit.googleapis.com/v1/accounts:signInWithPassword"
+        "?key={0}".format(api_key)
```
laurentS, line 81: Why not use an f-string here? I think it'd be more consistent with what you've done elsewhere.

mapswipe_workers/tests/integration/test_firebase_db_rules.py
```diff
@@ -0,0 +1,329 @@
...
+                f"observed, expected, get {endpoint} {user['displayName']}",
+            )
+            write_access = test_set_endpoint(user, path)
+            self.assertEqual(
+                write_access,
+                expected_access[i][1],
+                f"observed, expected, set {endpoint} {user['displayName']}",
+            )
+
+    def test_access_as_project_manager(self):
```
laurentS, line 286: I like how easy this makes it to add a new endpoint. I wonder if you could refactor most of the `test_access_as` functions by creating a new function like `def try_access_as(self, user: UserRecord, expected_access: List[List[str]]` which you then call 3 times for each type of user.

mapswipe_workers/tests/integration/test_firebase_db_rules.py
```diff
@@ -0,0 +1,329 @@
...
+        # sign in user with email and password to simulate app user
+        user = sign_in_with_email_and_password(
+            self.project_manager.email, f"{self.project_manager.display_name}_pw"
+        )
+
+        expected_access = [  # [read, write]
+            [True, False],  # public project query
```
laurentS, line 293: This is hairsplitting, and I don't think it really matters. I would have used a list of tuples here, because I think semantically it makes a bit more sense than a list of lists. But I wouldn't bother changing it now :)

mapswipe_workers/tests/integration/test_firebase_db_rules.py
laurentS: Actually another way to do this could be to derive 3 classes of tests, like `TestFirebaseDBRulesAsNormalUser` etc, and just modify what changes between them. At this point, I think this would require more rewriting than needed, but just an idea :)

Hagellach37: yeah, I had the same in mind, when copying the code for the third time. ;)

Basically the only thing that changes is the expected permission denied or granted. Everything else is the same.

Hagellach37:
Hey,
okay so this is it. :D I added a test for another private project the user is not a member of. The code refactoring will be left for the future. Then we should also take a closer look how code can be refactored between other tests, e.g. set up a team or project etc.

Issue #397, opened 2020-07-30
Limit tasks per user and project


Hagellach37:
Currently for each user there are these attributes in respect to their contributions: `projectContributionCount`, `taskContributionCount`, `groupContributionCount`.

In order to distribute tasks among team members more equally, there should be an option to limit the number of tasks a user can work on per project. This will require the following changes:

* [x] add `v2/teams/{teamId}/maxTasksPerUserPerProject`attribute during team creation in mapswipe back end
* [x] add counter to firebase functions for `v2/users/{uid}/contributions/{projectId}/taskContributionCount`, `../groupContributionCount` to user profile using `../projectId` as key (counter will be triggered when results are submitted to firebase)

Everything else will be handled by the app along these lines:
* once team member signs in, get `maxTasksPerUserPerProject`
* user wants to map --> check if `v2/teams/{MyTeamId}/maxTasksPerUserPerProject` < `v2/users/{uid}/contributions/{projectId}/taskContributionCount` --> query groups
* (the last statement could also be made as a part of the firebase db rules, but maybe this makes it too complicated and it will be enough to just check in the app)
* user uploads results to firebase once finishing the group --> this will trigger the firebase function to set the counters 

Open Questions:
* Should we define `maxTasksPerUser` on the team level or or the project level?

Hagellach37:
@laurentS again here I added some thoughts on how to implement that a user can only work on a defined number of tasks. Anything missing? what is your opinion on the `teams` vs. `project` for `maxTasksPerUser` attribute (or should it even be per user)?

Hagellach37:
we will go `maxTasksPerProject` defined on the project level. This can be set by the project manager during project creation

Hagellach37:
for now we will go with defining it on the team level. Let's see how often we have to use this feature and if we can do improvements then.

@laurentS so you can expect the information here: `v2/teams/{teamId}/maxTasksPerUserPerProject`

sorry for the back and forth

PR #401, opened 2020-08-04
Limit tasks per user

Hagellach37:
This refers to #397 

@laurentS do you think this is already enough from the perspective of the app?

This pull request introduces a major update of node.js from version `8` to version `10`. We have to do this, since otherwise we will not be able to introduce any changes to the existing firebase functions. I know that you once mentioned that this can have implications for all the code. What's a good way to test for it?

data structure in firebase looks like this for users:

Screenshot from 2020-08-04 19-36-44

### DIFF
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -47,6 +47,10 @@ exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{use
     const endTimeRef            = admin.database().ref('/v2/results/'+context.params.projectId+'/'+context.params.groupId+'/'+context.params.userId+'/endTime')
     const timeSpentMappingRef   = admin.database().ref('/v2/results/'+context.params.projectId+'/'+context.params.groupId+'/'+context.params.userId+'/timeSpentMappingRef')
 
+    // references for project based counters for tasks and groups
+    const projectTaskContributionCountRef      = admin.database().ref('/v2/users/'+context.params.userId+'/contributions/'+context.params.projectId+'/taskContributionCount')
+    const projectGroupContributionCountRef     = admin.database().ref('/v2/users/'+context.params.userId+'/contributions/'+context.params.projectId+'/groupContributionCount')
+
     // Counter for groups
     const groupFinishedCount = groupFinishedCountRef.transaction((currentCount) => {
         return currentCount + 1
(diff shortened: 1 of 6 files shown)


laurentS:
Hopping to a new major release of node.js in the middle of another PR doesn't look like the best idea to me, I think it's worth putting it in a PR by itself, even if you're only really changing 2 characters.
The ideal situation would obviously be to have a proper set of automated tests so you can feel confident about your upgrade, but it's not going to happen.
I would start by reading the [release notes](https://github.com/nodejs/node/releases/tag/v10.0.0) of node 10 quite carefully, focusing on breaking changes, and see if any of the new/removed/changed items apply to your code. This needs to be done by someone who knows the code base, so I guess you or Matthias? Then I would merge the dedicated PR, deploy to the dev environment, and let it run for a couple of days to check that nothing breaks.

laurentS:
About the data itself, is your idea that the app sends the `startTime` and `endTime` along with the resultts, and the backend calculates the other elements (`groupContributionCount` and `taskContributionCount`)? If so, I think we're good, when a user wants to map on a project, we can just check how many tasks they've contributed on it. The only open point in my mind is how to deal with the situation where they do multiple groups in a row: does the app get updates on the total from the backend (a bit more download to do), or does it need to update the total tasks mapped locally, and hope that the number matches the one in the backend?
I think the tricky bit might be around "continue mapping" where it might be easier to add counts locally, and if the user goes back to the menu, then we sync from the server again, otherwise I'll have to do a manual download of data, which may or may not be easy (I'm unsure yet). If you have any good insights here, I'm happy to hear them :)

Hagellach37:
Hey,
two good points. I can put the upgrade of node into a separate PR. However, saying this I upgraded node already to v10 on the dev server and my guess is that there is no way now to go back to v8... 

Regarding the data: the counters (`groupContributionCount`, `taskContributionCount`...) are updated by a firebase function, whenever new results are uploaded to firebase. This triggers a bunch of different counters, I just added two new counters. The counters for how many tasks a user contributed per project are put in the user section of firebase. 

So once a user finishes a group and uploads the results, this will increase the counter in his user profile. This is similar to the counter we already use when displaying the badge for the users and how many swipes/tasks they need to do untill they reach the next level.

Regarding the continue mapping part: during this part the data is not synced with firebase?

PR #407, opened 2020-08-11
Setup Travis secrets to enable access to Firebase

Matthias-Schaub:
With this Pull Request the Travis instance can now access the Firebase CI (firebase-test) project. The Docker container firebase-deploy now builds and runs successfully and tests as well.


.travis.yml
```diff
@@ -3,41 +3,24 @@
...
 
 install:
   - pip install --upgrade pip setuptools
   - pip install flake8 black
 
 script:
   - docker-compose up --build --detach firebase_deploy
+  - docker-compose up --build --detach postgres
   - docker-compose run mapswipe_workers python -m unittest discover --verbose --start-directory tests/unittests/
   #- docker-compose run mapswipe_workers python -m unittest discover --verbose --start-directory tests/integration/
```
Hagellach37, line 25: I think we should then also uncomment the intregration test

docs/source/dev_setup.md
```diff
@@ -136,13 +136,11 @@ For more information refer to the official [Reference on Cloud Function for Fire
 
 ## Travis Setup
 
-Configuration for travis setup is utilizing environment variables.
+A Travis instance is used to build MapSwipe Workers and run tests.
 
-One difference to production or development setup is that the service account key as json is stored in the environment variables `FIREBASE_CONFIG` as text. To make this work special characters as to be escaped first. This command will simply escape every character:
+For the configuration of Travis environment variables are used: https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings
```
Hagellach37, line 141: could you add here which environment variables are set in travis. I think we only set some and not all. E.g. `bing_imagery_api_key` or so is not set and doesn't need to be set.

mapswipe_workers/tests/unittests/test_firebase_connection.py
```diff
@@ -0,0 +1,24 @@
...
+class TestFirebaseConnection(unittest.TestCase):
+    def setUp(self):
+        self.fb_db = auth.firebaseDB()
+
+    def tearDown(self):
+        self.fb_db = None
+
+    def test_create_project(self):
+        ref = self.fb_db.reference("/v2/test")
+        ref.push("test")
```
Hagellach37, line 15: I think if you push something here, you should also remove it during tear down?

maybe there is a simpler test to check of the Firebase Connection is working.

mapswipe_workers/tests/unittests/test_firebase_connection.py
```diff
@@ -0,0 +1,24 @@
...
+        result = ref.get()
+        self.assertIsNotNone(result)
+        ref.set({})
```
Hagellach37, line 18: ah okay here you delete the data again...

mapswipe_workers/tests/unittests/test_firebase_connection.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class TestFirebaseConnection(unittest.TestCase):
+    def setUp(self):
+        self.fb_db = auth.firebaseDB()
+
+    def tearDown(self):
+        self.fb_db = None
+
+    def test_create_project(self):
```
Hagellach37, line 13: we should rename this function. Will this return something? Can we test, assert for a specific error in firebase? e.g. no connection or something like this?

e.g. defined here: https://firebase.google.com/docs/reference/admin/python/firebase_admin.exceptions

.gitignore
```diff
@@ -135,3 +132,5 @@ nginx.conf
 
 .DS_Store
 
+# Custom setup script
+setup-custom.sh
```
Hagellach37, line 136: when is this script used?

docs/source/dev_setup.md
Matthias-Schaub: Sure. I added more information on the travis setup to the docs.

mapswipe_workers/tests/unittests/test_firebase_connection.py
Matthias-Schaub: Yes there is a simpler test indeed. Now the tests just gets the keys at root with .get(shallow=True).

mapswipe_workers/tests/unittests/test_firebase_connection.py
Matthias-Schaub: Good idea. I created an issue for this: #408

.gitignore
Matthias-Schaub: At the moment I use this script to setup my environment for development.

PR #410, opened 2020-08-20
Tutorial new structure

Hagellach37:
Hey @laurentS ,
this should basically give you an overview what has changed in regard to the tutorial. All the data is still in firebase.

I basically renamed some of the attributes of the tutorial to allow for a simper set up. You can find a tutorial in the `dev` firebase here: `v2/projects/built_area_tutorial_solid_waste`

There is now also a web site to upload tutorial data. (this is not super important for now...)


firebase/database.rules.json
```diff
@@ -15,6 +15,8 @@
           ||
           (auth != null  && query.orderByChild == 'status' && query.equalTo == 'build_area_tutorial' && query.limitToFirst == 1)
           ||
+          (auth != null  && query.orderByChild == 'name' && query.equalTo == 'build_area_tutorial_solid_waste' && query.limitToFirst == 1)
```
Hagellach37, line 18: @laurentS this is how you could query the tutorial for now... I think we need to find a better way in the future to do this. we need to make sure that we can have multiple tutorials for built area. but for now it should be fine to query like this.

mapswipe_workers/sample_data/build_area/solid_waste_tutorial_screens.json
```diff
@@ -0,0 +1,98 @@
+{
```
Hagellach37, line 1: @laurentS this basically tells you for each screen the text strings. these are already in firebase as well. for simplicity it might be the best, if you change in firebase directly if needed.

laurentS: I think it would make it easier (at least on the app side) if the screens were 0-indexed, because with this format firebase returns an array of screens with the first element being `null`, which is a bit weird.

laurentS: Another tiny bug: xMax in the group was 117, when the tasks actually stop at 115. I've fixed in firebase, but you might want to check if there's a bug in the code that generated the data.

laurentS: We also need to think about how we're going to transition from the old tutorial format to the new one: the "old" app and the "new" one won't be able to use the same tutorial content, so there's probably going to be a period where either the old or the new version is broken for the tutorial. I can't think of a good way to handle this right now, unless maybe by creating a tutorial that has both the "categories" and "screens" in it, which we'd keep until the old version of the app is not around anymore (and we show an announcement asking users to update?)

Hagellach37:
@laurentS if you could do a last rough check here, this would be super great. I will merge into dev then to test things on the server.

mapswipe_workers/python_scripts/add_tutorial_id_to_projects.py
```diff
@@ -0,0 +1,44 @@
...
+from mapswipe_workers.definitions import logger
+
+
+def get_all_projects(project_type: int):
+    """Get the user ids from all users in Firebase DB."""
+
+    project_id_list = []
+    fb_db = firebaseDB()
+
+    status_list = ["active", "inactive"]
```
laurentS, line 12: Is it on purpose you're ignoring private projects here?

laurentS (approved):
I've left a bunch of comments and questions. No showstoppers, so I think you can merge right away if you don't want to wait :)

PR #411, opened 2020-09-10
add counters on project level

Hagellach37:


Hagellach37:
Hey @laurentS ,
in order to have counts per project we need to update the firebase functions and we have to upgrade to node.js 10...
There is no way around this. 

I would do this now to have at least one day to test things.

laurentS:
Ah, fun days! Let's do it! Do you need anything from me to update?

PR #412, opened 2020-09-10
Dev

Hagellach37:


postgres/scripts/changeGeomType.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE projects ALTER COLUMN geom SET DATA TYPE geometry;
```
Hagellach37, line 1: this is important. We need to run this once we deploy the changes to the production postgres db.

PR #413, opened 2020-09-11
Code style: Make code flake8 and black compatible. Fixes #327

### LINKED ISSUE #327: Code style: Make code flake8 and black compatible.
New and recently changed code is already flake8 and black compatible. Some code is not yet. In addition to flake8 and black type hints should be included.

After the code is compatible flake8 and black should be run in Travis as well.

gcaria:
I have found a problem about the compatibility of black and flake8.
I first run black on the repository, then went on to manually fix all the issues flake8 found (there were around 200 of them).
This is the code that this pull request contains.

The problem is that now running black on the code again will create around 150 `E501 line too long` flake8 warnings, since apparently black doesn't like many types of line wrapping. 
For example, this is a diff after running black one last time:
```
-        "No configuration for Sentry was found. Continue without Sentry"
-        " integration."
+        "No configuration for Sentry was found. Continue without Sentry" " integration."
```
Personally it seems that this edit is not necessary and the code reads better according to flake8 rather than black,
but it's a matter of taste.


Matthias-Schaub:
Hey gcaria,

great that you contribute. It would really nice to all the code in a consistent style.
Next time I would like you to ask to work an a branch which is based on the 'dev' branch and then make a pull request back to the 'dev' branch. That way we can make sure everything is working before merging it into master. We have a dedicated server on which the version of mapswipe from the dev branch is running. A bit on that is to find in the docs: https://mapswipe-workers.readthedocs.io/en/dev/contributing.html
(Although they need to be updated and rewritten in parts..)

I will edit this pull request to be merged with dev instead of master.

To your point with too long lines. This seems to be a bit odd because I can not replicate your described scenario.
It happend to me before that flake8 did not respect the setup.cfg config at the root of the project. In that it is redefined what the max line length of flake8 is to be (so to be compatible with black):
```ini
[flake8]
# max-line-lenght = 88 to conform with Black
max-line-length = 88
```

If you get a warning of flake8 where the max line lenght number is below 88 then the config has not been respected.
e.g. thats fine: ` E501 line too long (89 > 88 characters)`


Thanks for your work :)

Matthias-Schaub:
Actually I think it is not a good idea to merge this branch (based on the master) back into dev due to conflicts.
May I kindly ask you to do this again for the `dev` branch?

Also it seems that you touched non python files like deploy.sh and docker-compose.yaml and changed syntax to be invalid.

gcaria:
Hi @Matthias-Schaub 

my bad for not reading the contributing docs, should definitely have done that ! I wil try again in the `dev` branch.

Now I understand, since I was running both black and flake8 in the root directory of the project (`python-mapswipe-workers`) flake8 was not taking into account the `setup.cfg` config in `python-mapswipe-workers/mapswipe_workers`.

Since I saw python files also in other directories, such as:
```
python-mapswipe-workers/docs/source/conf.py
python-mapswipe-workers/postgres/scripts/v1_to_v2/generate_copy_to_csv.py
python-mapswipe-workers/test_config.py
```
I worked directly in the root directory. Was it planned that `setup.cfg` won't cover these files ?

Matthias-Schaub:
Yes, there are a many components in this repository, which can be confusing. You can think of each directory at root level as a separate component. `mapswipe_wokers` implements the logic of the back-end and is a python package. Hence setup.py, requirements.txt and setup.cfg are to be found in that directory. In theory mapswipe_workers can be installed without the other componets (But it would be of no use). So yes it is intentionally. 

Cheers

gcaria:
Great thanks ! I will close this and submit another PR

Issue #420, opened 2020-09-28
Firebase Functions: Project Contributor Count is broken


Hagellach37:
With the latest release of the firebase functions and the additional stats per project in the user profile it seems as if the counter for mappers per project is broken. It stays at 0 all the time.

This needs to be handled by firebase functions.

Matthias-Schaub:
Fixed in Pull Request #423

PR #423, opened 2020-10-08
Fix for "Project Contributor Count is broken" #420

Matthias-Schaub:
Firebase Functions
Refactor counter for first time user contribution to a project into seperate database trigger function.

Mapswipe Workers  Base Class:
Init contributorCount with 0.

### DIFF
--- a/firebase/functions/index.js
+++ b/firebase/functions/index.js
@@ -30,66 +30,26 @@ exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{use
 
 
     // Firebase Realtime Database references
+    const contributorCountRef  = admin.database().ref('/v2/projects/'+context.params.projectId+'/contributorCount')
+
     const groupFinishedCountRef = admin.database().ref('/v2/groups/'+context.params.projectId+'/'+context.params.groupId+'/finishedCount')
     const groupRequiredCountRef = admin.database().ref('/v2/groups/'+context.params.projectId+'/'+context.params.groupId+'/requiredCount')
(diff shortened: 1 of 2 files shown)


firebase/functions/index.js
```diff
@@ -163,6 +130,21 @@ exports.counter = functions.database.ref('/v2/results/{projectId}/{groupId}/{use
 })
 
 
+exports.counter_2 = functions.database.ref('/v2/users/{userId}/contributions/{projectId}/').onCreate((snapshot, context) => {
```
Hagellach37, line 133: I like the approach to put this into a separate function.

Maybe you can find a better name than `counter_2`. :)  e.g. `projectCounter` or something similar.

Matthias-Schaub:
I restructured (copy passed) code blocks in index.js. There are not that many real changes although git highlighted a lot.

Hagellach37 (approved):
Hey,
looks good to me. A general question: Do you think that there is further potential to untangle the first counter? it is doing a lot at the same time...

Maybe you can at least provide some comments for each of the counter function to describe the desired behaviour directly in this file. :)

Matthias-Schaub:
I agree that the functions could be simpler and structured better to enhance readability. I can spent some more time on the functions to improve readability and documentation. With this pull request I just wanted to fix the bug as soon as possible.

firebase/functions/index.js
Matthias-Schaub: Yeah counter_2 is not an elegant or meaningful name :D
I will think of something better and as you said include some more comments and maybe restructure more code if that improves readability. It took me some time to figure out again which attribute / db ref is depended on which other and what will execute when.

Hagellach37:
okay, agreed. Let's merge this one and address the other things separately. :)

PR #424, opened 2020-10-09
Updated tutorial text

poulamb:
Tracked-on: https://github.com/mapswipe/mapswipe/issues/418


Hagellach37:
Hey @poulamb ,
thanks for your contribution. I will merge this into dev now. To appear in the app it  will be required that we also deploy this to the master and production server. This might take some days, since I want to group some changes for the next release. But then your changes will become visible in the app.

Thanks,
Benni

PR #433, opened 2020-10-21
Feature smaller task size footprint

Hagellach37:
changes only apply to footprint project type. we use gzip and base64 encoding to reduce size of data for tasks in firebase.


mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/task.py
```diff
@@ -18,6 +18,9 @@ def __init__(self, group: object, featureId: int, featureGeometry: dict):
         super().__init__(group, taskId=task_id)
         self.geojson = featureGeometry
 
+        self.projectId = None
```
laurentS, line 21: is this removing project and group ID from all types of project? If so, I'd need to check that the app will work without it on `built_area` for instance

laurentS (changes requested):
Looks good, just need a few more comments, and confirm how widely you're removing the IDs.

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/task.py
laurentS: After testing on the client, this actually causes a problem, because it is converted to a literal `"projectId":None` string in the output. I can obviously parse the resulting string to remove it (and groupId), but it seems like a waste of bandwidth to carry things around to then throw them out. Maybe `del self.projectId` instead?

mapswipe_workers/mapswipe_workers/project_types/base/project.py
```diff
@@ -210,7 +213,14 @@ def save_to_firebase(self, project, groups, groupsOfTasks):
         c = 0
         for group_id, tasks_list in groupsOfTasks.items():
             c += 1
-            task_upload_dict[f"v2/tasks/{self.projectId}/{group_id}"] = tasks_list
+            if self.projectType in [ProjectType.FOOTPRINT.value]:
+                # we compress data for footprint project type
+                string_tasks = str(tasks_list).replace(" ", "")
```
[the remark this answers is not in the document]
laurentS: another issue is that `str(tasks_list)` does not produce valid JSON. I think `JSON.dumps(tasks_list)` might do the trick. The main difference is that `str()` uses single quotes everywhere, and the JSON parser in the app is not happy with it :cry:

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/task.py
Hagellach37: it's only removing groupId and projectId for footprint project types since this is in the `arbitrary_geometry/task.py`.

`arbitrary geometry` is the more general term of footprint I use for the back end.

Hagellach37:
@laurentS I will wait with merging this PR untill you tell me that things work out on the client as well.

Hagellach37:
@laurentS things on the client have been defined right? So I can merge this PR?

laurentS:
> @laurentS things on the client have been defined right? So I can merge this PR?

Yes, merge away! The PR is not merged on the client side, but will definitely be!

Issue #434, opened 2020-10-23
Project Creation: check for invalid groups 


Hagellach37:
During the project creation some groups contain 0 tasks. This results in a crash on the client and another back end error, when results should be transfered, e.g. `missing attribute 'results' for: -MIYSrmUZf_K-6jsPeZA/g569/iyx8Ae795aRKXTdkmL9PryIFVfm2, will skip this one`.

To avoid this we should check during project creation if a group is valid, e.g. check if the number of tasks per group is bigger than 0.

This also relates to #271

Hagellach37:
will be close once #436 merged into master

Issue #374, opened 2020-06-09
Project Creation: Allow Multiple Polygons as Input Extent for TMS Grid projects


Hagellach37:
Currently, one a single polygon geometry will be used as the input for the project creation workflow. 

There are several use cases for which it would be good to allow multiple polgons, e.g. having a project mapping several smaller islands. Currently, we would need to do a mapping project for each Island. It would be good, if we can have only a single project for this.

This requires changes in:
### Manager Dashboard: project creation form
* https://github.com/mapswipe/python-mapswipe-workers/blob/989eecb4aa501ffd7af47a0e01c0afe1cef4dcf8/manager_dashboard/manager_dashboard/js/forms.js#L145

### MapSwipe Workers: Project Geometry validation
*  https://github.com/mapswipe/python-mapswipe-workers/blob/989eecb4aa501ffd7af47a0e01c0afe1cef4dcf8/mapswipe_workers/mapswipe_workers/project_types/tile_map_service_grid/project.py#L60
* write a function `test_validate_geometries`, e.g. here: https://github.com/mapswipe/python-mapswipe-workers/tree/master/mapswipe_workers/tests/unittests

### MapSwipe Workers: project creation for tile map service grid projects
And now starts the complicated section... the important part here is that we need to make sure that the generated groups will not overlap each other. Hence, we need to make sure that each task can only be part of a single group.

![IMG_2140](https://user-images.githubusercontent.com/7045979/84148116-cbbc8980-aa5e-11ea-80a3-f627bfee2fad.JPG)

As displayed in the picture, this is related to how close the individual features are. If they are "distant enough" (whatever this exactly is), we can just run the existing workflow for each geometry and merge the results. But, if the geometries are "too close", this might lead to overlaps. In this context, "distant" and "close" are related to the zoom level of a project.

There might be several ways how we achieve this
* (A) apply a pre-processing step and make sure that all geometries are distant enough, if geometries are too close do some processing (e.g. buffer and when they intersect, create a union of both)
* (B) apply a post-processing step and make sure to remove all overlapping tasks from groups. E.g. make sure that whenever a task is part of multiple groups, it will be removed and remains only in a single group
* (C) don't care about overlapping tasks, this will lead to the fact that overlapping tasks will be worked on by more than the required number of volunteers and some tasks might be even worked on again by the same user (we should avoid C, ;))

Things to look at:
* adjust workflow of group creation. Group creation for TMS grid projects happens here:
[...]

Issue #271, opened 2020-01-15
Last screen only shows 3 cells


Hagellach37:
In the latest version (2.0.1) once you've swiped all the way to the right, the last screen often only has 3 tiles instead of the 6 I'd expect.

This has been discussed as a front end issue in https://github.com/mapswipe/mapswipe/issues/145 but then moved to the back end.

During project creation we need to make sure that groups have the correct size.

Hagellach37:
this should be tackled when working on #374

Hagellach37:
will be close once #436 merged into master

PR #435, opened 2020-10-24
Fix groups with no tasks

Hagellach37:
This PR fixes the grouping workflow for TMS projects ("normal" build area projects in the app)

* make sure that there are no groups without tasks #434 
* make sure that the group_x_size is always even, to make sure that a screen always shows 6 tiles #271 
* remove all overlapping groups which still occurred when using Multipolygons for project creation #374


mapswipe_workers/mapswipe_workers/utils/tile_grouping_functions.py
```diff
@@ -291,6 +299,53 @@ def groups_intersect():
             and (int(y_minB) <= int(y_max))
         )
 
+    def merge_groups():
```
laurentS, line 302: I would suggest making this a pure function, that takes 2 groups (or whatever you need) as arguments and returns a new group, and does not use any external variables. Right now, you're relying on context to give you the input data, and this tends to be very fragile, and therefore bug prone. If you move this function elsewhere in your code, or you change your code before or after this, you might change the meaning of a variable, and have to update the body of this function. You're breaking the "encapsulation" of this function in doing so, and it often leads to sad developers :)

mapswipe_workers/mapswipe_workers/utils/tile_grouping_functions.py
```diff
@@ -318,55 +373,17 @@ def groups_intersect():
 
             if groups_intersect():
```
laurentS, line 374: Same comment as above: I would make this a pure function (pure meaning it has no side effect and all the data it needs is clearly passed through arguments).

laurentS (approved):
Looks ok to me, aside from a comment about making functions better encapsulated. This is mostly to limit the risk of bugs creeping in later on.
I'm also wondering why you create groups and then filter them to remove odd-length ones. Is there no way to make sure that their size is even when initially creating them? It seems like it would be less expensive computationally.

Hagellach37:
Hey @laurentS,
one quick comment on why there is a need to filter the odd length groups. In the first place we create groups that have an even length. But: if groups overlap by a single tile, we just merge these two groups into a single group. But this merge operation can result in an odd group length. That's why we have to check again after merging groups.

I will address you other comments. Thanks for the suggestions. :)

Issue #455, opened 2021-03-20
Simplify Firebase Functions and move project progress calculation to python backend


Hagellach37:
This issue is a result of the discussion that we had in another issue #449, about firebase und postgres stats not showing the same values.

Basically the outcome is that Firebase functions should be idempotent.

Firebase functions should do the following things that need to happen in realtime in an idempotent way:

* [x] set `finishedCount` and `requiredCount` for a group in Firebase database
* [x] set mapping stats in the user profile (this will be used to check which groups a user already worked on)
    * [x] set `contributions` for a group in a project
    * [x] set total  `taskContributionCount` for project
    * [x]  set `taskContributionCount` for project

The python backend should do the following things, which are hard to implement in an idempotent way in Firebase:
* [x] update project `progress` in Firebase database based on results in postgres database
* [x] `contributorCount` for a project in Firebase database based on results in postgres database

Hagellach37:
@laurentS which user stats have to be set for the app in its current state?

I think that the app only uses the `taskContributionCount` attribute. Can you confirm this?

How should we handle this in general? Should these values (and potential other user stats values) be updated by a firebase function, by the python backend or directly in the app?

Hagellach37:
@laurentS and another question: which attribute is used to check against the maxNumberOfTasks per project in private projects?

laurentS:
So looking through the code, I'm seeing:
- `taskContributionCount` is used in the "more" screen, to show the progress bar and level badge
- `groupContributionCount` and `projectContributionCount` are not used at this point. They are just set to `0` when a user account is created, but that's it.
- I need to document how the `maxNumberOfTasks` is used, but in short, we compare it with `v2/users/<userId>/contributions/<projectId>/taskContributionCount` (see [here](https://github.com/mapswipe/mapswipe/blob/3bb8bb762dbe00e63521815643dbb8924084fe64/src/shared/views/ProjectView.js#L454))

laurentS:
And yes, I think these should be updated in "realtime", or at least within a minute or so, otherwise users will probably feel like something is not working.

Issue #449, opened 2021-01-30
Firebase and postgresql based stats do not always agree


laurentS:
I'm writing this ticket mostly to keep track of a complex bug we spent hours tracking with @Hagellach37 .
In some projects, at some periods of time, there is a difference between the progress reported by the app (from firebase data) and the one reported on the mapswipe.org/data.html page (from postgresql data).
The progress shown in the app is calculated from the data in firebase, by the firebase functions, which are called each time an app uploads results for a group. The `finishedCount` attribute on a group shows how many such results have been reported by the app, and should increase by 1 every time a user completes a group. `requiredCount` is updated at the same time, but down from the initial number of users required to review each task (typically 3 or 5). These numbers are updated at the same time as the progress percentage, so are a good proxy to work out the source of the problem.
This bug is about how this `finishedCount` and `requiredCount` differ from the actual number of results seen in firebase and postgresql.
The answer seems to be that firebase functions are "guaranteed to be invoked at least once". From [the docs](https://cloud.google.com/functions/docs/concepts/exec#execution_guarantees):
> Background functions are invoked at least once. This is because of the asynchronous nature of handling events, in which there is no caller that waits for the response. The system might, in rare circumstances, invoke a background function more than once in order to ensure delivery of the event. If a background function invocation fails with an error, it will not be invoked again unless retries on failure are enabled for that function.

So we need to modify the functions so that they stop execution in case they are called more than once.
An example solution is proposed in [this stackoverflow question](https://stackoverflow.com/a/54712252/1138710).

Extra github comments from other people mentioning the issue, and google staff replies:
- https://github.com/firebase/functions-samples/issues/105
- https://github.com/firebase/firebase-functions/issues/315#issuecomment-419810580
- https://github.com/firebase/firebase-functions/issues/62#issuecomment-419810195

Hagellach37:
Hey,
I see a few options how this could be done in firebase functions.

What we could for instance write the userId to the group object in firebase, e.g. create a structure like this:

```
groupA:
    verificationCount: 5,  // this will be set during project creation
    users:  // this will be set by a firebase function that gets triggered when a result is uploaded
      userA: true,
      userB: true,
   fininishedCount: 2,
   
```

Based on this new structure, we could re-calculate the `finsihedCount` and `requiredCount` attributes, based on the `verificationCount` and the number of userIds we get for the users attribute.

When a function would be trigerred again this would not change the counters of a group then.

laurentS:
[...]

PR #456, opened 2021-03-22
Fix realtime progress

Hagellach37:
This moves some functionality from firebase functions to the python backend as described here #455 . We now calculate progress and number of contributors for a project only in the backend, since there have been several issues with these calculations using firebase functions (#449).

We still calculate the number of projects, groups and tasks a user has worked on with firebase functions, since this is currently used by the leaderboard (projects, groups, tasks) and in the app (tasks)

In the user contributions we only log the groupId and do not add further information on starttime and endtime. This information can be derived from the postgres database if needed.


firebase/functions/index.js
```diff
@@ -27,118 +38,116 @@ exports.resultCounter = functions.database.ref('/v2/results/{projectId}/{groupId
...
+                return {
+                    userContribution: userContributionRef.child(context.params.groupId).set(data),
+                    groupUsers: groupUsersRef.child(context.params.userId).set(true),
+                    totalTaskContributionCount: totalTaskContributionCountRef.transaction((currentCount) => {return currentCount + numberOfTasks}),
+                    taskContributionCount: taskContributionCountRef.transaction((currentCount) => {return currentCount + numberOfTasks})
+                }
             }
         })
-    promises.push(contributionTime)
+    promises.push(updateValues.userContribution)
```
laurentS, line 70: is this not going to crash if `updateValues` finds existing results and returns `null` (line 75/52)? This line here would try to read an attribute of `null`, which I'd expect to :boom:

firebase/functions/index.js
```diff
@@ -27,118 +38,116 @@ exports.resultCounter = functions.database.ref('/v2/results/{projectId}/{groupId
...
-        })
-        .then((numberOfTasks) => {
-            return userRef.child('taskContributionCount').transaction((currentCount) => {
-                return currentCount + numberOfTasks
-            })
-        })
-    promises.push(totalTaskContributionCount)
-
-    // Counter for User contributions on a distict Project
-    const groupContributionCount = userContributionRef.child('groupContributionCount').transaction((currentCount) => {
```
laurentS, line 54: I think we were a bit too hasty in removing this number, it's used in the leaderboard :fearful:

firebase/functions/index.js
```diff
@@ -5,12 +5,23 @@ const functions = require('firebase-functions')
 const admin = require('firebase-admin')
 admin.initializeApp()
 
-// Increments or decrements various counters of User and Group once new reults are pushed.
+
+// Calculate number of users who finished a group and
```
laurentS, line 9: Can you add an overview explanation of how the `requiredCount` and all get updated via `groupUsers`? I think it's not immediately obvious when reading the code below.

laurentS (commented):
Leaving some comments here for traceability in parallel to our chat.

firebase/functions/index.js
Hagellach37: okay, so this will then also have implications on the code in general. What's your feeling should these stats by set in realtime (by a firebase function) or updated just once in a while (e.g. every 15 minutes by the backend)?

My feeling is, that since these attributes are not used in the app, it doesn't have to be realtime.

mapswipe_workers/python_scripts/remove_unused_user_attributes.py
```diff
@@ -0,0 +1,42 @@
+from firebase_admin import auth
+
+from mapswipe_workers.auth import firebaseDB
+from mapswipe_workers.definitions import logger
+
+
+def remove_unused_user_attributes(uid):
+    """Set username in Firebase DB based on auth.display_name for user id."""
```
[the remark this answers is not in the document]
Hagellach37: as pointed out below, these values are used by the leaderboard and thus we will not delete them.

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
```diff
@@ -397,6 +416,17 @@ def get_per_project_statistics(project_id: str, project_info: pd.Series) -> dict
...
 
+        # update progress and contributors in firebase
+        set_progress_in_firebase(
+            project_id=project_id,
+            progress=int(100*project_stats_by_date_df["cum_progress"][-1])
+        )
+
+        set_contributors_in_firebase(
+            project_id=project_id,
+            contributors=int(project_stats_by_date_df["cum_number_of_users"][-1])
```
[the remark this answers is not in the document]
Hagellach37: I renamed this one to `contributor_count` and added a bit more comments in another file to make it easier to understand how attribute names used in firebase and python match.

firebase/functions/index.js
Hagellach37: I updated this part of the code and check if `updateValues` returns null

firebase/functions/index.js
Hagellach37: okay some further thoughts: We could for now keep it in Firebase and use the functions we already have. When re-designing the user profile section we can have this discussion again and decide which user stats we would like to provide and how they should be calculated.

When re-designing the user profile we can also think about this will affect the leaderboard. Ideally we can bring those two things closer together and maybe we do not need a leaderboard when we improve the user profile.

firebase/functions/index.js
```diff
@@ -27,118 +42,126 @@ exports.resultCounter = functions.database.ref('/v2/results/{projectId}/{groupId
...
+    */
+    updateValues = groupUsersRef.child(context.params.userId).once('value')
         .then((dataSnapshot) => {
             if (dataSnapshot.exists()) {
                 console.log('group contribution exists already. user: '+context.params.userId+' project: '+context.params.projectId+' group: '+context.params.groupId)
                 return null
             }
             else {
+
                 const data = {
```
Hagellach37, line 60: @laurentS another question here: I think we do not have to set this, since we are currently not further using this information and it just takes up space in the user profile...

firebase/functions/index.js
```diff
@@ -27,118 +45,140 @@ exports.resultCounter = functions.database.ref('/v2/results/{projectId}/{groupId
...
                 console.log('group contribution exists already. user: '+context.params.userId+' project: '+context.params.projectId+' group: '+context.params.groupId)
                 return null
             }
             else {
-                const data = {
-                    'startTime': result['startTime'],
-                    'endTime': result['endTime']
+                const numberOfTasks = Object.keys( result['results'] ).length
+                return {
+                    userContribution: userContributionRef.child(context.params.groupId).set(true),
```
Hagellach37, line 64: @laurentS do you see any problem here?

We are not using the data that is set in the user contributions in Firebase as far as I know. The app just checks for the groupId in `v2/users/{userId}/contributions/{projectId}/{groupId}`. Is this correct?

laurentS: yes, none of that is used in the app

laurentS (approved):
looks all good!

Issue #457, opened 2021-03-22
docker container structure: split jobs for project creation and result processing


Hagellach37:
Currently we use a single docker container that performs three steps one after the other:

* project creation / tutorial creation
* transfer results from firebase to postgres
* generate stats

It would be better to split this into two separate containers / jobs. This will help us to make the back end more reliable. E.g. an error during project creation and tutorial creation will not affect the transfer of results and generate stats part and vice versa.

* [x] adjust docker compose file
* [x] adjust run function in mapswipe_workers.py

PR #458, opened 2021-03-23
split workers into two separate docker containers

Hagellach37:
The idea behind this is described in #457 

Basically, we just split the jobs into two docker containers.


laurentS (approved):
Looks great to me!

Issue #478, opened 2021-04-19
Firebase transactions get aborted when a lot of mapping happens


Hagellach37:
It might be better to not use a transaction here, but do it the "normal" way. So download the results and once they go inserted into the postgres db, update the refs in firebase.

Hagellach37:
Hey @laurentS @Matthias-Schaub ,
it would be great to get an opinion from you on this. :)

Hagellach37:
Together with @laurentS we discussed some ideas:

## The old version
* just for reference: it takes 7 seconds to transfer results for 21 groups
* around 3 seconds to get user_ids from postgres
* around 4 seconds to insert results in postgres table

con:
* transaction may get aborted when a lot of users map at the same time

## Transactions on the group level
* we could keep the logic to use transaction, but apply it one level below
* we would use transaction not for all results of a project, but for the groups

con:
* this is rather slow, when just using "normal" queries
* transfering results for 21 groups took around 80 seconds
* this is too slow in situations when many results get added

## Do not use transaction
* keep the logic on the project level
* this is how we did it before mapswipe backend version 2, e.g. check here: https://github.com/mapswipe/python-mapswipe-workers/blob/e0116879df1c55d3184c04c7893e9b17add645ca/mapswipe_workers/basic/BaseFunctions.py#L746
* this taskes around 7 seconds and reaches a similar performance than the existing workflow

con:
* using firebase transaction might be a bit safer (trying to avoid deleting data we don't want to delete)

PR #480, opened 2021-04-19
Fix transfer results transaction aborted

Hagellach37:
This refers to #478 

Here we try to improve the transfer of results from firebase to postgres. 

Basically, we keep the existing logic, but don't use a Firebase transaction function anymore. 

* [x] check for potential exceptions and handle


mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -55,24 +42,92 @@ def transfer(current_results):
...
-            logger.exception(
-                f"{project_id}: Firebase transaction for "
-                f"transfering results failed to commit"
-            )
-            sentry.capture_exception()
+        # It is important here that we first insert results into postgres
+        # and then delete these results from Firebase.
+        # In case something goes wrong during the insert, results in Firebase
+        # will not get deleted.
+        delete_results_from_firebase(project_id, results)
```
Matthias-Schaub, line 103: Here we have to be careful. This should only be executed if `save_results_to_postgres` has been successful. Currently we hope that this function (and the psycopg library) will raise an exception when inserting results to the db has not been successful. I do not know if there is need for other checks and how they would look like.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -55,24 +42,92 @@ def transfer(current_results):
...
+
+    fb_db = auth.firebaseDB()
+
+    # we will use a multi-location update to delete the entries
+    # therefore we create a dict with the items we want to delete
+    data = {}
+    for group_id, users in results.items():
+        for user_id, result in users.items():
+            key = f"{group_id}/{user_id}"
+            data[key] = None
```
Matthias-Schaub, line 124: Does setting the reference (`fb_db.reference(f"v2/results/{project_id}/")`) with a dict containing this key (`"{group_id}/{user_id}"`) work as expected?

I would build a nested dict as follows:
```python
    for group_id, users in results.items():
        data[group_id] = {}
        for user_id in users.keys():        
             data[group_id][user_is] = None
```

But it is a while since I worked with Firebase.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -55,24 +42,92 @@ def transfer(current_results):
...
+    Save results into an in-memory file.
+    Copy the results to postgres.
+    Delete results in firebase.
+
+    We are NOT using a Firebase transaction functions here anymore.
+    This has caused problems, in situations where a lot of mappers are
+    uploading results to Firebase at the same time. Basically, this is
+    due to the behaviour of Firebase Transaction function:
+
+        "If another client writes to this location
```
laurentS, line 64: Can I suggest adding a link to the source? This page is not what you quoted, but also has helpful info on transactions https://firebase.google.com/docs/firestore/manage-data/transactions#transaction_failure

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
laurentS: To second what Matthias said, you want to make sure that you delete results only if there is no error when saving to postgres. You could use the python exception construct:
```python
try:
    save_results_to_postgres()
except SomePGError:
    log("failed")
else:
    delete_results_from_firebase()  # this is only run if there is no exception in the try block
```
See https://docs.python.org/3/tutorial/errors.html for more details. This only works if any problem in the postgres execution raises an exception, but I think this is a fairly safe assumption (reading the docs never hurts ;)

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -55,24 +42,92 @@ def transfer(current_results):
...
-                f"{project_id}: Firebase transaction for "
-                f"transfering results failed to commit"
-            )
-            sentry.capture_exception()
+        # It is important here that we first insert results into postgres
+        # and then delete these results from Firebase.
+        # In case something goes wrong during the insert, results in Firebase
+        # will not get deleted.
+        delete_results_from_firebase(project_id, results)
+        del fb_db
```
laurentS, line 104: Any special reason you delete `fb_db` here? it will get deleted automatically next line, when it becomes out of scope.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -55,24 +42,92 @@ def transfer(current_results):
...
+    # we will use a multi-location update to delete the entries
+    # therefore we create a dict with the items we want to delete
+    data = {}
+    for group_id, users in results.items():
+        for user_id, result in users.items():
+            key = f"{group_id}/{user_id}"
+            data[key] = None
 
+    results_ref = fb_db.reference(f"v2/results/{project_id}/")
+    results_ref.update(data)
```
laurentS, line 127: What happens here if there are new results uploaded since data was read in firebase? Are they not impacted by this `update`?

laurentS: I'm thinking it would probably be useful to write a unit test for this:
- create fake result data in firebase
- read it with the function above
- simulate a user uploading a result for a different group
- delete_results_from_firebase() and check that the extra result has not been removed/altered

laurentS (changes requested):
I think there's a couple of change to make to help limit data loss. I don't know enough about how firebase works with updates, so maybe worth checking the docs a bit more (unless you know the answers to the points below from both Matthias and myself).
The key problem you're trying to solve here is write conflicts, when multiple users write to firebase at the same time, and that's hard to test for. The best I can think of would be to do a bit of load testing using something like https://locust.io/, and simulate N writes within a given time period, and check that each of them ends up in postgres.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: Hey, @Matthias-Schaub 
it works. So basically the first part in that dict is the path. That's why it looks a bit strange.

This is also covered by the tests, in which we check that the results are actually deleted from firebase.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: Hey, 
yep this is a very good comment. I changed it based on your comments.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -9,37 +9,24 @@
...
-
-    fb_db = auth.firebaseDB()
+    """Transfer results for one project after the other.
 
+    Will only trigger the transfer of results for projects
+    that are defined in the postgres database.
+    Will not transfer results for tutorials and
+    for projects which are not set up in postgres.
+    """
     if not project_id_list:
```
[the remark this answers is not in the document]
Hagellach37: adjusted

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: I added a link to the Firebase python sdk where I copied the statement from.

You link refers to FireStore, which is another thing. :)

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: no, removed

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: Hey,
thanks for this comment. I added a new test for this case and check that the results that are uploaded only after the transfer results process has started, are not deleted.

Hagellach37:
Hey @laurentS  @Matthias-Schaub 
thanks for your thoughtful comments. This is exactly what I was hoping for.

Based on your suggestions I adjusted the code and added new tests as well.

Maybe one of you can take a look again if this satisfies you.

mapswipe_workers/tests/integration/test_transfer_results.py
```diff
@@ -81,6 +86,52 @@ def test_user_not_in_postgres(self):
         result = pg_db.retr_query(sql_query)
         self.assertIsNotNone(result)
 
+    def test_additional_results_are_not_deleted(self):
```
Hagellach37, line 89: @laurentS this is the test that you suggested.

mapswipe_workers/tests/integration/test_transfer_results_invalid_results.py
```diff
@@ -0,0 +1,51 @@
...
+class TestTranserResultsProject(unittest.TestCase):
+    def setUp(self):
+        self.project_id = set_up.create_test_project(
+            "tile_map_service_grid", "build_area", results=False
+        )
+
+    def tearDown(self):
+        tear_down.delete_test_data(self.project_id)
+
+    def test_invalid_task_in_result(self):
```
Hagellach37, line 21: @laurentS here I test for an psycopg2 related error during the transfer of the results

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
laurentS: Hmm, just to be on the safe side, would it make sense to add `except Exception:` before the `else:` block here? The way you do it now, if any exception that is not a `ForeignKeyViolation` is raised in the `try` block, it won't get handled, and the process will crash. If you add another handler, you can get some cleaner logging, and keep going. I'm not sure what makes more sense.

mapswipe_workers/tests/integration/test_transfer_results.py
laurentS: I don't know enough about how this works, apologies if I'm asking dumb stuff: the initial results are apparently created in group 115, so this test does not check for group clash, right? Would it be worth adding an extra test where you have additional results in the same group 115 during the transfer process? Basically the same as this one, only changing the groupId in the extra results.

mapswipe_workers/tests/integration/test_transfer_results_invalid_results.py
```diff
@@ -0,0 +1,51 @@
...
+
+        fb_db = auth.firebaseDB()
+        ref = fb_db.reference(f"/v2/results/{self.project_id}")
+        ref.set(test_data)
+
+        transfer_results.transfer_results(project_id_list=[self.project_id])
+
+        fb_db = auth.firebaseDB()
+        ref = fb_db.reference("v2/results/{0}".format(self.project_id))
+        self.assertDictEqual(ref.get(shallow=True), {"g115": True})
```
laurentS, line 47: Maybe I am being paranoid, but this is not actually checking that the results are still there, just verifying the presence of the key. Would checking something like  `fb_db.reference("test_build_area/results/18-156540-142783").get == 1` make sense?

mapswipe_workers/tests/integration/test_transfer_results_many_results.py
```diff
@@ -0,0 +1,39 @@
...
+        fb_db = auth.firebaseDB()
+        ref = fb_db.reference(f"v2/results/{self.project_id}")
+        self.assertIsNone(ref.get())
+
+        pg_db = auth.postgresDB()
+        sql_query = (
+            f"SELECT * "
+            f"FROM results "
+            f"WHERE project_id = '{self.project_id}' "
+            f"AND user_id = '{self.project_id}'"
```
laurentS, line 32: Is the user_id meant to be the same as the project_id?

Hagellach37: yes, it's a bit confusing but in this case they are really the same.

mapswipe_workers/tests/integration/test_transfer_results.py
Hagellach37: I added another test for this. So we have two test cases now:

1. a new result for a different group in the results
2. a new result for an existing group in the results

mapswipe_workers/tests/integration/test_transfer_results_invalid_results.py
Hagellach37: Hey,
I adjusted this part and hope that it is now "paranoid save". :D

I check for the number of results, rather than for a specific value.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
Hagellach37: yes, you are right. I added some new lines.

Hagellach37:
Hey @laurentS ,
anything more that I should do here? Or can I merge and deploy to the dev server?

laurentS:
> anything more that I should do here? Or can I merge and deploy to the dev server?

Nothing I can think of. But we should probably try to load test the dev server at some point

Issue #498, opened 2021-11-23
check if we get the changeset comments from osm api


ElJocho:

ElJocho:
@Hagellach37 we get a comment if it is set, e.g. "RDC-Congo, Nord Kivu #hotosm-task-122, source=Bing" for https://www.openstreetmap.org/api/0.6/changeset/14372614.json

-> should we add that?

Hagellach37:
yes, let's add it for now. But maybe let's check how "big" the geojson files get then.

Issue #496, opened 2021-11-19
Add extra info for footprint project tasks


ElJocho:
- [x] query extra info from ohsome inluding userid, last _edited, osm_feature_id, version(how many times it got changed), source (imagery source) -> in metadata: lastedit, version, changesetid, 
- [x] this info should reach postgres via geojson but not firebase -> remove it in base/project.py/save_to_firebase
- [x] get userid and name from https://www.openstreetmap.org/api/0.6/changeset/changeset_id_we_got_from_metadata and write into properties as well -> to avoid extra queries save already queried changesetids and there response in temporary dict. ->>>> /api/0.6/changesets -> query for more changesets
- [x] adjust the way the groups are formatted, see local todo

-> clean up this issue and outline order

Hagellach37:
could you also check if we want to add the hot tasking manager project id here as well or only at the project level.

ElJocho:
-> input in tasks: Geojson Feature old screen, center and reference included in the geojson feature but not in property section. (take them from property section and put them out of it if its a tutorial, only tutorials have this attribute)

geojson variable -> add a comment that is only the geometry of the feature

Issue #495, opened 2021-11-19
Remove special handling for google Task


ElJocho:

ElJocho:
@Hagellach37 we get a comment if it is set, e.g. "RDC-Congo, Nord Kivu #hotosm-task-122, source=Bing"

-> should be add that?

PR #500, opened 2021-12-06
Add more properties based on osm and ohsome data to geojson and push them into postgres

ElJocho:
Solves:
- #495
- #498
- #496


mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/grouping_functions.py
```diff
@@ -1,8 +1,6 @@
 import argparse
```
Hagellach37, line 1: I think we can remove all argparse related things here as well. This is just old overhead we don't need anymore. I guess it somehow comes from the times when this script was used directly from the command line. We don't do this anymore and if we want to test the function, we will write a test. So it should be safe to remove it here.

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/grouping_functions.py
```diff
@@ -65,50 +61,22 @@ def group_input_geometries(input_geometries_file, group_size):
...
         except KeyError:
-            # todo: change this and call feature_geoms -> geojson
-            groups[group_id_string] = {
-                "feature_ids": [],
-                "feature_geometries": [],
-                "center_points": [],
-                "reference": [],
-                "screen": [],
-            }
+            groups[group_id_string] = {"feature_ids": [], "feature": []}
```
Hagellach37, line 64: to be strict we would use `features` here, right? as this will be a list of more than one feature.

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/grouping_functions.py
```diff
@@ -65,50 +61,22 @@ def group_input_geometries(input_geometries_file, group_size):
...
-            groups[group_id_string]["center_points"].append(None)
-        else:
-            groups[group_id_string]["center_points"].append([center_x, center_y])
 
         # this is relevant for the tutorial
-        reference = feature.GetFieldAsDouble("reference")
-        screen = feature.GetFieldAsDouble("screen")
-        groups[group_id_string]["reference"].append(reference)
-        groups[group_id_string]["screen"].append(screen)
+        if "screen" in feature.keys():
```
Hagellach37, line 71: I'm not sure if this is going to work. The feature here is the geojson feature, right? Did you test this explicitly with a file that contains these columns in the geojson file?

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/grouping_functions.py
```diff
@@ -65,50 +61,22 @@ def group_input_geometries(input_geometries_file, group_size):
...
     return groups
 
 
-# todo: groups should have this output formatting
-# reference is correct answer for tutorial,
-# if not tutorial reference = None
-# {"g100":[{"task_id":1, "screen": 1, "reference":0,
-# "geojson":{"valid geojson FeatureCollection Object"}}]}
 ########################################################################################
 if __name__ == "__main__":
```
Hagellach37, line 81: as mentioned above, just remove this here as well

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/group.py
```diff
@@ -8,14 +8,7 @@ class Group(BaseGroup):
...
 
-    def create_tasks(
-        self,
-        feature_ids: List,
-        feature_geometries: List,
-        center_points: List,
-        reference: List,
-        screen: List,
-    ) -> None:
+    def create_tasks(self, feature_ids: List, feature: List,) -> None:
```
Hagellach37, line 11: as mentioned below we could rename `feature` into `features`

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/task.py
```diff
@@ -7,43 +7,30 @@
...
-
+        self.geojson = feature["geometry"]
+        self.properties = feature["properties"]
         # only tasks that are part of a tutorial need this
-        if screen:
-            self.screen = screen
-            self.reference = reference
+        if "screen" in feature["properties"].keys():
+            self.screen = feature["properties"]["screen"]
+            self.reference = feature["properties"]["screen"]
```
Hagellach37, line 25: this should be `feature["properties"]["reference"]` and not "screen"

mapswipe_workers/mapswipe_workers/project_types/base/project.py
```diff
@@ -230,7 +235,8 @@ def save_to_firebase(self, project, groups, groupsOfTasks):
                 # since the tasks hold geometries their storage size
                 # can get quite big otherwise
                 if self.projectType in [ProjectType.FOOTPRINT.value]:
-                    # todo: remove extra information from geojson
+                    # removing properties from each task
+                    tasks_list = [task.pop("properties", None) for task in tasks_list]
```
Hagellach37, line 239: I'm not sure if this works as expected. What is written into the list? would it be enough just to have the loop and not write the output back to the list?

mapswipe_workers/mapswipe_workers/project_types/base/project.py
```diff
@@ -587,14 +593,20 @@ def create_tasks_txt_file(self, groupsOfTasks):
                 # these common attributes don't need to be written
                 # to the project_type_specifics since they are
                 # already stored in separate columns
-                common_attributes = ["projectId", "groupId", "taskId", "geometry"]
+                common_attributes = [
+                    "projectId",
+                    "groupId",
+                    "taskId",
+                    "geometry",
+                    "geojson",
```
Hagellach37, line 601: perfect, that put this here. I think so far we stored the geometry twice. So this is much better.

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -13,7 +18,74 @@ def geojsonToFeatureCollection(geojson):
...
+        "@lastEdit": 0,
+        "@osmId": 0,
+        "@version": 0,
+        "source": 0,
+        "comment": 0,
+    }
+
+    for feature in json["features"]:
+        new_properties = {}
+        for attribute in wanted_rows:
```
Hagellach37, line 63: maybe just to make this code easier to read you could think about putting the following check into a separate function. But you can decide what you think works better.

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -13,7 +18,74 @@ def geojsonToFeatureCollection(geojson):
...
+                        attribute
+                    ]
+                else:
+                    new_properties[attribute.replace("@", "")] = feature["properties"][
+                        "tags"
+                    ][attribute]
+            except KeyError:
+                missing_rows[attribute] += 1
+        changeset_id = new_properties["changesetId"]
+        if changeset_id not in changeset_results.keys():
```
Hagellach37, line 76: here you check if you queried the changeset information already, right? maybe add a comment to the code so that this is more clear

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -25,12 +97,17 @@ def ohsome(request: dict, area: str, properties=None):
...
         logger.warning(response.json())
         raise CustomError(err)
     else:
         logger.info("Query succesfull.")
-    return response.json()
+
+    response = response.json()
+
+    if properties:
+        remove_noise_and_add_user_info(response)
```
Hagellach37, line 112: this function should return something, or not?

mapswipe_workers/tests/integration/test_create_footprint_project.py
```diff
@@ -23,8 +23,9 @@ def tearDown(self):
 
     def test_create_footprint_project(self):
```
Hagellach37, line 24: here we should now probably also check if the tasks contain the attributes we would like to have?

not sure what is the best option. maybe you add a new test function and just check the tasks?

mapswipe_workers/mapswipe_workers/project_types/base/project.py
ElJocho: yup, thats a big no no

Hagellach37 (commented):
Hey @ElJocho ,
good work and nice to see that we can now have the additional information for the tasks. Could you check my comments and resolve once you worked on them or you think that we should not address them?

Most things look fine, but probably it would be good to add a test to check if the additional attributes are written in the postgres db as we expect it.

mapswipe_workers/mapswipe_workers/project_types/arbitrary_geometry/grouping_functions.py
ElJocho: true, the whole part is unnecessary, no error because it is not executed and screen and reference are already in properties by default

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -13,7 +18,74 @@ def geojsonToFeatureCollection(geojson):
     return geojson
 
 
-def ohsome(request: dict, area: str, properties=None):
+def query_osm(changeset_id: int):
+    """Get data from changesetId."""
+    url = OSM_API_LINK + f"changeset/{changeset_id}.json"
+    valid_response = False
+    times_timed_out = 0
+    while not valid_response:
```
Hagellach37, line 26: looks interesting. So you want to try again after a timeout, correct?

maybe check this blog post how to achieve it a bit nicer: https://www.peterbe.com/plog/best-practice-with-retries-with-requests

mapswipe_workers/mapswipe_workers/utils/api_calls.py
ElJocho: good call for readability; worked and not caused any errors because dict is immutable, so it does not have to be returned

mapswipe_workers/mapswipe_workers/utils/api_calls.py
ElJocho: done

mapswipe_workers/mapswipe_workers/project_types/base/project.py
```diff
@@ -121,8 +121,13 @@ def save_project(self):
             del group["tasks"]
             groups[group["groupId"]] = group
         del project["groups"]
+
         project.pop("inputGeometries", None)
         project.pop("validInputGeometries", None)
+        project.pop("inputType", None)
```
Hagellach37, line 127: I think I overlooked this before. Why do you delete these project attributes? I think we can just keep them. They might be useful information that the people setting up projects can use later, e.g. to find out which projects they created already for which tasking manager projects

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -13,7 +27,70 @@ def geojsonToFeatureCollection(geojson):
...
+            attribute
+        ]
+    return new_properties
+
+
+def remove_noise_and_add_user_info(json: dict) -> dict:
+    """Delete unwanted information from properties."""
+    logger.info("starting filtering and adding extra info")
+
+    wanted_attributes = ["@changesetId", "@lastEdit", "@osmId", "@version", "source"]
```
Hagellach37, line 60: any reason that here you don't have `comment`, but you have it in the missing_rows dictionary below?

mapswipe_workers/tests/integration/test_create_footprint_project.py
ElJocho: I added one test which counts how many usernames made it into project_specifics->properties->username. If above 0 made it we can guess that ohsomeinfos and changesets arrived at the db. 

Do you think that is enough or do you have more ideas for tests?

mapswipe_workers/mapswipe_workers/project_types/base/project.py
ElJocho: keeping them now :+1:

mapswipe_workers/tests/integration/test_create_footprint_project.py
Hagellach37: perfect, yes I think this should be enough.

Hagellach37 (approved):
Hey @ElJocho ,
this looks good now. Let's merge. :)

Issue #501, opened 2021-12-06
[manager dashboard] infinite "be patient" when no project image provided


Hagellach37:

ElJocho:
Is a project image mandatory, or do we have a default picture?

Hagellach37:
I think we should make it mandatory. There is no default image.

PR #503, opened 2021-12-08
catch missing image 

ElJocho:
fixes:
- #501

additionally fixed double point before gif


ElJocho:
what other attributes should be mandatory; do we want to address this now or open a future issue?

Hagellach37 (approved):
Thanks! Tested and worked well.

Issue #502, opened 2021-12-07
OSM Changeset comments are not saved


Hagellach37:
We try to save the changeset comments when downloading OSM geometries. However, we need to adjust where we look for these comments. And probably to be more precise, actually we do not want to save the comment, but the hashtags. E.g. check here: https://www.openstreetmap.org/api/0.6/changeset/54128350.json

We need to check in `item["elements"][i]["tags"]["host"]`.

```
{
    "version":"0.6",
    "generator":"CGImap 0.8.6 (1731532 spike-08.openstreetmap.org)",
    "copyright":"OpenStreetMap and contributors",
    "attribution":"http://www.openstreetmap.org/copyright",
    "license":"http://opendatacommons.org/licenses/odbl/1-0/",
    "elements":[{
     "type":"changeset",
  "id":54128350,
"created_at":"2017-11-27T17:51:29Z",
"closed_at":"2017-11-27T17:51:31Z",
"open":false,
"user":"marco443",
"uid":1886646,
"minlat":18.3552561,
"minlon":-66.8353373,
"maxlat":18.3573810,
"maxlon":-66.8324753,
"comments_count":0,
"changes_count":111,
"tags":{"changesets_count":"27389","hashtags":"#hotosm-project-3696;#HurricaneMaria;#PuertoRico","imagery_used":"Custom (https://{switch:services,server}.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x})","locale":"fr","host":"http://www.openstreetmap.org/edit","created_by":"iD 2.5.1","comment":"#hotosm-project-3696 #HurricaneMaria #PuertoRico"}}
]}
```

ElJocho:
weird we do already look in tags, where the comment should be stored, changing that to hashtag should be sufficient.
Could it be that you tested this on an older commit, I only added comment in a recent commit here: https://github.com/mapswipe/python-mapswipe-workers/pull/500/commits/eda27f591ceae0e1e22129299bb21f68452f0a8f?

```python:    
if attribute != "comment":
      new_properties[attribute.replace("@", "")] = feature["properties"][attribute]
else:
      new_properties[attribute.replace("@", "")] = feature["properties"]["tags"][attribute]
```

PR #504, opened 2021-12-08
Changed comment to hashtags and moved to right section

ElJocho:
solves:
- #502


Hagellach37 (approved):
Thanks for correcting this. I've tested it and just changed a bit. I removed the source attribute, as it is not so widely used as expected and added the created_by attribute instead. this will be more interesting as we might be able to see differences between users that used JOSM and users that used iD.

PR #505, opened 2021-12-09
bundle changesets

ElJocho:
Bundles are requested without threading.


mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -27,33 +29,76 @@ def geojsonToFeatureCollection(geojson: dict) -> dict:
...
     """Get data from changesetId."""
-    url = OSM_API_LINK + f"changeset/{changeset_id}.json"
+    id_string = ""
+    for id in changeset_ids:
+        id_string += f"{id},"
+
+    id_string = id_string[:-1]
 
+    url = OSM_API_LINK + f"changesets?changesets={id_string}"
+    logger.info(url)
```
Hagellach37, line 51: Hey,
let's remove the logging of the url.
But it would be nice to add a log after the query and logging the progress. Something like `finished query 1/15`

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -27,33 +29,76 @@ def geojsonToFeatureCollection(geojson: dict) -> dict:
...
+        for tag in changeset.iter("tag"):
+            if tag.attrib["k"] == "hashtags":
+                hashtags = tag.attrib["v"]
+            if tag.attrib["k"] == "created_by":
+                created_by = tag.attrib["v"]
+
+        osm_results[int(id)] = {
+            "username": username,
+            "userid": userid,
+            "hashtags": hashtags,
```
Hagellach37, line 75: sorry for the back end forth on this one here. Finally I come to the conclusion that we should use `comment` instead of `hashtags`. 

I've realized that the `hashtags` are only set if the user used the iD editor. If you use JOSM this is not set. However, both editors put the hashtags also in the comment.

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -64,30 +109,25 @@ def remove_noise_and_add_user_info(json: dict) -> dict:
...
+    )
+    chunk_list = chunks(list(osm_results.keys()), 100)
+    for subset in chunk_list:
+        query_osm(subset)
+
+    for feature in json["features"]:
+        changeset = osm_results[feature["properties"]["changesetId"]]
+        feature["properties"]["username"] = changeset["username"]
+        feature["properties"]["userid"] = changeset["userid"]
+        feature["properties"]["hashtags"] = changeset["hashtags"]
```
Hagellach37, line 128: see above. Let's adjust and use `comment` instead of `hashtags`

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -27,33 +29,76 @@ def geojsonToFeatureCollection(geojson: dict) -> dict:
     return geojson
 
 
-def query_osm(changeset_id: int):
+osm_results = {}
```
Hagellach37, line 32: I'm not sure if there is another option to setting this here. From my point of view it might make more sense to define this in the query_osm function and then return `osm_results` at the end of the function.

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -27,33 +29,76 @@ def geojsonToFeatureCollection(geojson: dict) -> dict:
...
         raise CustomError(err)
-    response = response.json()["elements"][0]
-    return response
+    tree = ElementTree.fromstring(response.content)
+
+    for changeset in tree.iter("changeset"):
+        id = changeset.attrib["id"]
+        username = changeset.attrib["user"]
+        userid = changeset.attrib["uid"]
+        hashtags = created_by = None
```
Hagellach37, line 65: that's interesting, this works?

Hagellach37 (changes requested):
Hey @ElJocho ,
most things are clear and look good. Can you just check what other options there are when it comes to the `osm_results`?

mapswipe_workers/mapswipe_workers/utils/api_calls.py
ElJocho: ye ^^ = resolves from right to left

ElJocho:
> Hey @ElJocho , most things are clear and look good. Can you just check what other options there are when it comes to the `osm_results`?

changed to changeset_results

Hagellach37:
Hey @ElJocho ,
great! :) I fixed a small bug about how the progress was displayed and also needed to check for `"` in the changeset comment. But now everything should be fine and is ready to get merged. You want to do that?

Hagellach37 (dismissed):

Hagellach37:
ah just realizied that I introduced a bug...

mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -27,33 +29,74 @@ def geojsonToFeatureCollection(geojson: dict) -> dict:
...
+        changeset_results[int(id)] = {
+            "username": username,
+            "userid": userid,
+            "comment": comment,
+            "created_by": created_by,
+        }
+    return changeset_results
+
+
+def add_to_properties(attribute: str, feature: dict, new_properties: dict):
```
Hagellach37, line 79: it looks as if this function is not used anymore.

Hagellach37:
Hey @ElJocho ,
so maybe it's best if you check this one quickly again and remove the `add_to_properties` function if if is not needed.

Can you please also check if there is a better place to check for `"` in the changeset_comment, created_by and username?

mapswipe_workers/mapswipe_workers/utils/api_calls.py
ElJocho: ye you are right, sneaked back in on merge

ElJocho:
okay i am positive that it is now on a presentable state :).

Hagellach37:
yes, looks much cleaner now. Great and thanks a lot. :)

Hagellach37 (approved):

PR #509, opened 2021-12-29
Osm api to osmcha

ElJocho:
-> query everything that is not in osmcha api in osm api as bulk downloads.

### DIFF
--- a/mapswipe_workers/mapswipe_workers/utils/api_calls.py
+++ b/mapswipe_workers/mapswipe_workers/utils/api_calls.py
@@ -1,9 +1,12 @@
+from xml.etree import ElementTree
+
 import requests
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
 from mapswipe_workers.definitions import (
     OHSOME_API_LINK,
+    OSM_API_LINK,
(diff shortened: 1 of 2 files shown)


mapswipe_workers/mapswipe_workers/utils/api_calls.py
```diff
@@ -22,12 +24,16 @@ def remove_troublesome_chars(string: str):
...
 
-def retry_get(url, retries=3, timeout=4):
+def retry_get(url, retries=3, timeout=4, to_osmcha: bool = False):
     """Retry a query for a variable amount of tries."""
     retry = Retry(total=retries)
     with requests.Session() as session:
         session.mount("https://", HTTPAdapter(max_retries=retry))
-        return session.get(url, timeout=timeout)
+        if to_osmcha:
+            headers = {"Authorization": OSMCHA_API_KEY}
```
Hagellach37, line 33: this seems not to work for me locally. Maybe I'm not using the right format here. What would you expect here?

Hagellach37:
Alright,
I think I've got in running now. I was a bit confused by the way how the OSM Cha Token is passed via the API. But now things are fine and working. I will merge this PR. 
Thanks for your work! :)

PR #510, opened 2022-01-04
Export changes

ElJocho:


ElJocho:
fixed old tasks in dev postgres via 
```sql
UPDATE tasks
SET project_type_specifics = jsonb_set(project_type_specifics::jsonb, '{properties, userid}', translate((project_type_specifics->'properties'->'userid')::text, '"', ' ')::jsonb)::json
```

Hagellach37:
Hey @ElJocho ,
this goes into the right direction, but I think we need to make sure to handle the different project types individually or better handle the cases when the additional attributes for the tasks are not present.

Ideally the code would stay the same for the normal project type. I will add some comments in the code directly and then we can discuss further ideas.

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
```diff
@@ -118,11 +118,19 @@ def get_tasks(filename: str, project_id: str) -> pd.DataFrame:
...
         sql_query = sql.SQL(
             """
-            COPY (
-                SELECT project_id, group_id, task_id, ST_AsText(geom) as geom
-                FROM tasks
-                WHERE project_id = {}
-            ) TO STDOUT WITH CSV HEADER
+COPY (
+    SELECT project_id, group_id, task_id, ST_AsText(geom) as geom,
+     (project_type_specifics->'properties'->'osmId')::text as osmId,
```
Hagellach37, line 123: interesting, this seems to work also for normal project types and the value for osmId will just be `NULL` in such situations. However, this will look a bit strange in the csv file.

Ideally we can check the project type here and then either use the old query or in case it is a "building footprint" project then we use your query.

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
```diff
@@ -304,7 +312,22 @@ def get_agg_results_by_task_id(
 
     # add task geometry using left join
     agg_results_df = results_by_task_id_df.merge(
```
Hagellach37, line 314: same as above it might be good to check the project type first and then join the columns.

Hagellach37: or you just add a check here if the additional columns actually exist in the `task_df` dataframe. Then we don't need the project type information.

ElJocho: would a simple df.dropna(how='all', axis=1, inplace=True) suffice? -> drops all empty columns

Hagellach37: yes, that would be a good option. I think ideal would be that we only drop the additional columns that are empty. And always keep the original ones, e.g. 1_count, 2_count etc.

mapswipe_workers/mapswipe_workers/generate_stats/project_stats.py
Hagellach37: or maybe you can check in the write_sql_to_gzipped_csv function if you can simply drop the empty columns, so that they don't show up in the csv file as well.

ElJocho:
testable project e.g. -MtPHP6M-2rDuEww2AUN

PR #521, opened 2022-02-10
Setup osm oauth backend functions

laurentS:
This PR sets up firebase functions that allow logging in to the app using an OSM accounts.

There are some open points left:
- [x] improving security of the token function with the use of the cookie
- [x] configuring the REDIRECT_URI which should be different for prod/dev apps
- [x] configuring the OSM_API_URL based on which deployment is being done
- [x] confirming that upgrading node from 10 to 12 does not break existing functions
- [ ] removing excessive logging once we are comfortable that the workflow is reasonably reliable, and definitely before pushing it to production, as this currently lists user tokens

This PR must be reviewed along the client side one: https://github.com/mapswipe/mapswipe/pull/618


firebase/functions/osm_auth.js
```diff
@@ -0,0 +1,238 @@
...
+    // user account (linked to the OSM account) exists.
+    const userCreationTask = admin
+        .auth()
+        .updateUser(uid, {
+            displayName: displayName,
+        })
+        .catch(error => {
+            // If user does not exists we create it.
+            if (error.code === 'auth/user-not-found') {
+                return admin.auth().createUser({
```
Hagellach37, line 224: at what point in the workflow does the app write things to the user profile and when the firebase functions?

I think that currently the firebase functions do this when you login with OSM and the app does it when you log in with email + password. But I think that the app writes more data to firebase, e.g. the app saves the username to the user profile, the firebase functions don't do this. The app further displays a banner after successful login saying "welcome...", this is also not the case for the OSM login.

we could either check that both write the same information: https://github.com/mapswipe/mapswipe/blob/master/src/shared/views/Login.js#L203-L231

Hagellach37 (commented):
Hey @laurentS ,
I've checked you changes and left some comments. I think that the overall workflow is more or less clear to me. Nicely done.

I have some questions around how the behaviour of the functions compares to the one of the app when you log in. It might be best to have a quick chat if my comments are not clear.

firebase/functions/osm_auth.js
Hagellach37: another quick question here: https://dev-mapswipe.firebaseio.com/v2/OSMAccessToken/osm:undefined

this is probably an artifact from testing? how can the osm user id be `undefined`. I think that in such a case we will not create an acccount.

firebase/README.md
```diff
@@ -18,3 +18,21 @@ Now you are inside the docker container and can login to firebase. You need to i
 Finally you can deploy your changes for cloud functions and database rules individually.
 * `firebase deploy --only functions`
 * `firebase deploy --only database:rules`
+
+## Notes on OAuth (OSM login)
+
+The main notes are in the app repository (/docs/osm_login.md).
+
+Some specifics about the related functions:
+ - get a service-account.json file from firebase which allows the OAuth functions to access the database and call
```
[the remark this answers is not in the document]
laurentS: I used a separate one while testing to avoid messing things up. I think we should be able to reuse the current one, we just need to double check it for permissions. I think this just needs executing functions and accessing the database (and opening non-firebase urls, but we get that from being on the blaze plan, apparently).

firebase/README.md
```diff
@@ -18,3 +18,21 @@ Now you are inside the docker container and can login to firebase. You need to i
...
+   external URLs (this last point only works on a firebase Blaze plan)
+- Before deploying, set the required firebase config values with:
+- `firebase functions:config:set osm.client_id="" osm.client_secret=""`
```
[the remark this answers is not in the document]
laurentS: done!

firebase/README.md
```diff
@@ -18,3 +18,21 @@ Now you are inside the docker container and can login to firebase. You need to i
...
+- Deploy the functions as explained above
+
+- Expose the functions publicly through firebase hosting
```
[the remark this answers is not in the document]
laurentS: Good catch, I missed that config!

firebase/database.rules.json
```diff
@@ -137,6 +137,12 @@
           "created",
           "teamId"
         ]
+      },
+      "OSMAccessToken": {
```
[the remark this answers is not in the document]
laurentS: We can do that too. The only thing to check is whether we ever access `v2/<userId>` for anything that is not done by the user, for instance stats or anything like this. Though I guess we could put a rule on that to make sure only a user has access to their own key (if RTDB allows this).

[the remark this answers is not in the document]
Hagellach37: ah you are right! the leaderboard reads the user data from firebase... and currently we only restrict who can write things in the user profile, but everyone can read everything.

So let's keep it like this. the user profile should contain public information. Privacy relevant things like osm auth tokens go under their own path.

firebase/functions/osm_auth.js
laurentS: Good catch on the `undefined` user. I've put a check in the function so it will error out in that case.
I have the same issue with the popup not being shown. I was convinced I had seen it when I first tested :thinking: 
As for updating the username, it looks like I missed the bit of code that does it.

nginx/nginx.conf
```diff
@@ -1,6 +1,6 @@
 server {
```
Hagellach37, line 1: I'm  not really sure why this file has changed. I guess it should have not.

docker-compose.yaml
```diff
@@ -198,6 +198,17 @@ services:
         environment:
             FIREBASE_TOKEN: '${FIREBASE_TOKEN}'
             FIREBASE_DB: '${FIREBASE_DB}'
-        command: sh -c "firebase use $FIREBASE_DB && firebase deploy --token $FIREBASE_TOKEN --only functions,database:rules"
+            OSM_OAUTH_REDIRECT_URI: '${OSM_OAUTH_REDIRECT_URI}'
+            OSM_OAUTH_API_URL: '${OSM_OAUTH_API_URL}'
+            OSM_OAUTH_CLIENT_ID: '${OSM_OAUTH_CLIENT_ID}'
+            OSM_OAUTH_CLIENT_SECRET: '${OSM_OAUTH_CLIENT_SECRET}'
+        command: >-
```
Hagellach37, line 205: here we do not deploy `hosting`. Is this on purpose?

firebase/README.md
```diff
@@ -15,6 +15,51 @@ Then run the container interactively and open a bash shell.
 Now you are inside the docker container and can login to firebase. You need to insert an authorization code into the terminal during that process.
 * `firebase login --no-localhost`
 
-Finally you can deploy your changes for cloud functions and database rules individually.
-* `firebase deploy --only functions`
+Finally you can deploy your changes for cloud functions and database rules individually. Hosting must be done as well to
+expose the authentication functions publicly.
+* `firebase deploy --only functions,hosting`
```
Hagellach37, line 20: Is there a reason why you deploy these separately?

Hagellach37:
Will merge this now. CI for firebase database rules passes now. CI still failed, but for another reason when creating a footprint project, because the ohsome api was down at the moment.

PR #537, opened 2022-05-02
adjust sql query to make sure to select all invalid results to delete

Hagellach37:
the previous solution didn't remove all invalid tasks. There could have been a situation where the task_id was twice listed in the results. E.g. one time for a valid result and another time for an invalid result. In such a situation the invalid result would not have been removed. 

This triggered many error messages in Sentry because the transfer of results from firebase to postgres is performed every 2 minutes (and failed every two minutes for the project)


laurentS (approved):
Looks fine to me, but maybe worth adding a line or two of explanation about why this is needed, and what the logic is for removing results.

Hagellach37:
Hey @laurentS ,
thanks. I've added a comment to explain what we do. There is also already a general explanation of this in `transfer_results.py`line 107.

PR #540, opened 2022-05-31
add command to update tileserver api key for a project

Hagellach37:


mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -406,6 +406,36 @@ def run_delete_project(project_id, project_ids):
...
+    cls=PythonLiteralOption,
+    default="[]",
+    help=(
+        "Set api key for multiple projects. "
+        "Provide project id strings as a list: "
+        """'["project_a", "project_b"]'"""
+    ),
+)
+@click.option("--api-key", help="the new api key to use", type=str, required=True)
+def run_set_tileserver_api_key(project_id, project_ids, api_key):
```
laurentS, line 427:
```suggestion
def run_set_tileserver_api_key(project_id: str, project_ids: Optional[list[str]], api_key: str) -> None:
```
You'll need to `from typing import Optional` as well

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
```diff
@@ -406,6 +406,36 @@ def run_delete_project(project_id, project_ids):
...
+    """Change the imagery API key for a project in Firebase."""
+    if not project_ids and not project_id:
+        click.echo("Missing argument")
+        return None
+    elif not project_ids:
+        project_ids = [project_id]
+
+    for project_id in project_ids:
+        update_data.set_tileserver_api_key(project_id, api_key)
```
laurentS, line 436: Is there some easy check you could run on the `api_key` argument to help catch typos and stuff? Maybe check the length if it's known, or ideally, run a test request with the new api key to check that the server accepts it.

laurentS (approved):
Just some ideas to make it a bit more robust.

mapswipe_workers/mapswipe_workers/mapswipe_workers.py
Hagellach37: I think this is a bit more complex. Here both parameters are optional `project_id`  and `project_ids`. But one of the two needs to be defied. we check for this at the beginning of the function.

PR #542, opened 2022-06-07
fix github actions code style and other issues

Hagellach37:


Hagellach37:
Hey @ElJocho ,

do you want to give this a quick look? It fixes the issue you had with the tests and also adds some more things to make the tests more reliable in Github Actions setting.

PR #544, opened 2022-06-16
Add UserGroup in Manager Dashbaord

frozenhelium:
### Firebase
- Update database rule for user groups
### Manager Dashboard
- Add new page to manage user groups
- Add User Group navigation menu in all existing pages

## TODO
- [x] Optimize query to fetch users for selected group
- [x] Add sanitization for user group key


manager_dashboard/manager_dashboard/usergroup.html
```diff
@@ -0,0 +1,201 @@
...
+          </div>
+          <div class="modal-body">
+            <div class="row">
+              <div class="col-md-12">
+                <h3>User group Information</h3>
+                <form class="form-style-7" id="usergroup-form">
+                  <ul>
+                    <li>
+                      <label for="usergroup-name">Name</label>
+                      <input type="text" name="usergroupName" id="usergroup-name" maxlength="30" required>
```
laurentS, line 161: Should this value be cleaned/validated before being accepted? What would happen if a user input something like `../broken.value/` or `group/child`?

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
...
+  $('#usergroup-description').val('');
+})
+
+function submitNewUserGroup() {
+  var nameInput = $('#usergroup-name');
+  var name = (nameInput.val() || '').trim();
+  var descriptionInput = $('#usergroup-description');
+  var description = (descriptionInput.val() || '').trim();
+
+  if (name && name.length <= 50) {
```
laurentS, line 12: In the other file, the group name max length is 30 chars?

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
...
+    $('#add-usergroup-modal-loading-message-container').removeClass('hidden');
+    nameInput.attr('disabled', true);
+    descriptionInput.attr('disabled', true);
+
+    firebase.database().ref('v2/userGroups/' + name).once("value", function(snapshot) {
```
laurentS, line 17: Please do some validation on the value input by user before using it like this.

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
...
+          userGroupName: name,
+          description,
+        };
+        firebase.database().ref().update(updates, function(error) {
+          $('#add-usergroup-modal-loading-message-container').addClass('hidden');
+          nameInput.attr('disabled', false);
+          descriptionInput.attr('disabled', false);
+
+          if (error) {
+            alert('Failed to add new user');
```
laurentS, line 35: Is there a way to provide a more helpful error message to the user, so that they can correct the problem? I'm not sure what errors are possible, and what would cause them, but it's always nice to give users a hint about what they could do.

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
...
+    firebase.database().ref('v2/userGroups/' + name).once("value", function(snapshot) {
+      if (snapshot.exists()) {
+        $('#add-usergroup-modal-loading-message-container').addClass('hidden');
+        nameInput.attr('disabled', false);
+        descriptionInput.attr('disabled', false);
+        alert('A group with this name already exists');
+      } else {
+        var updates = {};
+        updates['/v2/userGroups/' + name] = {
+          userGroupName: name,
```
laurentS, line 26: Do we need to duplicate the `name` here? Also, can the `userGroupName` key be shortened to `name`, for instance? I'm thinking of data volume and transfer cost in firebase :moneybag:

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
...
+            window.location.reload();
+          }
+        });
+      }
+    });
+  } else {
+    nameInput.attr('disabled', false);
+    descriptionInput.attr('disabled', false);
+    $('#add-usergroup-modal-loading-message-container').addClass('hidden');
+    alert('User group name is required and should be less than 50 characters');
```
laurentS, line 48: Same comment as above, you've limited this to 30 elsewhere.

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
...
+  const userGroup = JSON.parse(this.getAttribute('data-usergroup'));
+  var groupId = this.id;
+  var usersLength = Object.keys(userGroup.users || {}).length;
+
+  var tableRef = $("#groupMembersTable").DataTable();
+  // remove all existing rows
+  tableRef.clear();
+  var rows = []
+
+  var ref = firebase.database().ref("v2/users").orderByChild(`userGroups/${userGroup.userGroupName}`).limitToLast(usersLength);
```
laurentS, line 96: This query might return a LOT of data. Each `users` object includes the whole list of their past contributions. For a group with hundreds or thousands of active users, you could be downloading several MB of data here. Is there a way to limit what data is being fetched?

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,115 @@
+$("#add-usergroup-modal").on('show.bs.modal', function() {
+  $('#usergroup-name').val('');
+  $('#usergroup-description').val('');
+})
+
+function submitNewUserGroup() {
```
laurentS, line 6: Can you add a short comment to each function to explicit what it does?

firebase/database.rules.json
```diff
@@ -143,6 +143,10 @@
           ".read": "auth.uid === $uid",
           ".write": "auth.uid === $uid"
         }
+      },
+      "userGroups": {
+        ".read": true,
```
laurentS, line 148: Can this be made readable only by logged in users? Something like `.read: auth != null`? (I have not tested this)

manager_dashboard/manager_dashboard/js/usergroup-management.js
laurentS: Also, there are more than 50k users in production, so you might want to create an index on that child key to keep the query reasonably fast.

laurentS (changes requested):

manager_dashboard/manager_dashboard/usergroup.html
frozenhelium: Will add sanitization

manager_dashboard/manager_dashboard/js/usergroup-management.js
frozenhelium: Sure, I'm figuring out ways to optimize this query

manager_dashboard/manager_dashboard/js/usergroup-management.js
frozenhelium: Done!

manager_dashboard/manager_dashboard/js/usergroup-management.js
frozenhelium: Done!

manager_dashboard/manager_dashboard/js/usergroup-management.js
frozenhelium: Done!

manager_dashboard/manager_dashboard/js/usergroup-management.js
frozenhelium: Done!

firebase/database.rules.json
frozenhelium: Done!

laurentS:
@frozenhelium Can I suggest you push regular commits instead of overwriting everything on each push now that you've removed the extraneous files? It's very hard to track the changes you made like this, and it's a bit of a waste of time having to go through the entire PR every time :)

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,150 @@
...
+    alert('User group description is should be less than 100 characters');
+  } else {
+    showUserGroupLoading();
+
+    var key;
+    try {
+      key = encodeStringToKey(name);
+    } catch {
+      // This should never occur
+      alert('Failed to encode name into a key, please make sure that the user group name does not contain any invalid characters.');
```
laurentS, line 47: Your typical user probably won't know what characters are not valid for firebase keys :) Maybe you could add a list of what is allowed (or not, depending on what's shortest).

manager_dashboard/manager_dashboard/js/usergroup-management.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    if (key) {
+      firebase.database().ref('v2/userGroups/' + key).once("value", function(snapshot) {
+        if (snapshot.exists()) {
+          hideUserGroupLoading();
+          alert('A group with this name already exists');
+        } else {
+          var updates = {};
+          updates['/v2/userGroups/' + key] = {
```
laurentS, line 57: Thinking again about my comment during the call, I think using the user provided name is very likely to cause a lot of sadness down the line :sob: If a group is renamed (very likely, given how many users ask us to rename their user account), you would need to atomically change the key in multiple places within firebase. Not impossible, but cumbersome.

The other problem is that in the app, you get the list of groups as a `.once()` operation, without subscribing to updates. So if I open the list of groups, an admin renames the group I'm about to join, and then I join the group with the old name (because the app won't know about the rename unless I close and reopen the screen), you have a bug :bug: 

I think you could structure the data so that each group has a unique ID (generated in code, ideally a non sequential value so that enumeration is harder), and you should be able to check if the name already exists with something like `ref.child('userGroups').orderByChild('name').equalTo('MyNewGroup')`. What do you think? 
(this won't prevent simultaneous inserts of a single value if 2 admins create a group with the same name at the exact same time, but given that we have a handful of admins but thousands of users, it's a lot less likely to happen)
Firebase database does not seem to have anything like unique constraints, so I guess this is the best we can do.

laurentS (changes requested):
I added a few more points. I think the groupId should be a stable id. (I'm fighting with this exact problem right now trying to do stats on data from github, where the repository names can change. You end up with records for repo "mapswipe/oldthing" and others for "mapswipe/newthing" but they're really the same. It's very hard to fix this. I think you'll remove a ton of potential problems if you make sure you have a stable (ie: that does not change through the life of the record) id for each group)

manager_dashboard/manager_dashboard/js/usergroup-management.js
frozenhelium: @laurentS the concerns you have completely makes sense, let discuss it further on that part. I've thought about the query that you mentioned.
```
ref.child('userGroups').orderByChild('name').equalTo('MyNewGroup')
```
The major problem with this one would be that we cannot make the case insensitive check. So 'MyUserGroup' would be treated differently from 'Myusergroup'.

frozenhelium:
@laurentS I've used a bit different approach to ensure the unique group name. I've also found out some incorrect usage of promises in the firebase function. I've fixed them as well. Please do check them out as well.

CC: @Hagellach37

PR #562, opened 2022-09-06
UserGroup analytics

frozenhelium:
## Deployment helper steps

- Setup
  - Update .env using example.env
    - Make sure https certificates are created for used `NGINX_` domains.
- Build `docker-compose build`
- Update all the containers `docker-compose up`
- Apply new SQL changes to the database.
    ```bash
    docker-compose exec -T postgres bash -c 'psql -U $POSTGRES_USER $POSTGRES_DB' < ./postgres/scripts/userGroups.sql
    ```
- Apply migration on Django part
    - `docker-compose exec django ./manage.py migrate`
- Generate Organization Data for existing databases.
    - `docker-compose exec django ./manage.py update_organization_name`
- Generate/Update aggregated data
    - For Production:
        - Precalculate the data and then import it.
            ```bash
            cat ~/pre-calcuated-aggregated-data.csv | docker-compose exec -T django ./manage.py import_aggregated_data
            ```
    - For Dev/Staging:
        - This will also run automatically every day through django-schedule-task docker container.
            ```bash
            docker-compose exec django ./manage.py update_aggregated_data
            ```
Postgresql Optimization Setups (Optional)
- `ANALYZE`: It will optimize the query planner.
   https://www.postgresql.org/docs/current/sql-analyze.html
   https://andreigridnev.com/blog/2016-04-01-analyze-reindex-vacuum-in-postgresql/

- Aggregate calculation
    - Storage: NVME SSD
    - Time: 1h 41m


django/apps/existing_database/models.py
```diff
@@ -0,0 +1,213 @@
...
+    look_for = models.CharField(max_length=-1, blank=True, null=True)
+    name = models.CharField(max_length=-1, blank=True, null=True)
+    progress = models.IntegerField(blank=True, null=True)
+    project_details = models.CharField(max_length=-1, blank=True, null=True)
+    project_type = models.IntegerField(blank=True, null=True)
+    required_results = models.IntegerField(blank=True, null=True)
+    result_count = models.IntegerField(blank=True, null=True)
+    status = models.CharField(max_length=-1, blank=True, null=True)
+    verification_number = models.IntegerField(blank=True, null=True)
+    # Database uses JSON instead of JSONB (not supported by django)
```
Hagellach37, line 90: should we use JSONB then?

django/apps/existing_database/models.py
```diff
@@ -0,0 +1,213 @@
...
+    project_type_specifics = models.TextField(blank=True, null=True)
+    organization_name = models.CharField(max_length=-1, null=True, blank=True)
+
+    class Meta:
+        managed = False
+        db_table = "projects"
+        unique_together = (("project_id", "organization"),)
```
Hagellach37, line 97: just checking for clarification that I get this right. What do you define here? I don't understand why you also have `organization` here.

django/apps/existing_database/models.py
```diff
@@ -0,0 +1,213 @@
...
+    def __str__(self):
+        return f"P:{self.project_id}-G:{self.group_id}-T:{self.task_id}"
+
+
+class Result(models.Model):
+    # NOTE: Primary Key: project_id, group_id, tasks_id, user_id
+    project = models.ForeignKey(Project, models.DO_NOTHING, primary_key=True)
+    group_id = models.CharField(max_length=-1, primary_key=True)
+    task_id = models.CharField(max_length=-1, primary_key=True)
+    user = models.ForeignKey(User, models.DO_NOTHING, primary_key=True)
```
Hagellach37, line 155: here you define `project` and `user` based on ForeignKeys. Why are you not doing the same for group_id and task_id? In the current implementation this would mean that I can have a result, which defines a group_id which is not present in the groups table. Or do I miss something?

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+def get_contributor_time_stats() -> List[ContributorTimeType]:
+    COMMUNITY_DASHBOARD_STATS_QUERY = f"""
+        WITH
+        user_data AS (
+            SELECT
+                MAX(R.start_time) as start_time,
+                MAX(R.end_time) as end_time,
+                R.timestamp as timestamp
+            From {Result._meta.db_table} R
+            GROUP BY timestamp
```
Hagellach37, line 100: I want to double check here, which table this query is running against. can we group by timestamp or can it happen that there we get the same timestamp several times?

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+        aggregate_results = cursor.fetchall()
+    list_data = []
+    for data in aggregate_results:
+        list_data.append(
+            ContributorTimeType(total_time=round(data[1] / 60) or 0, task_date=data[0])
+        )
+    return list_data
+
+
+def get_project_type_query() -> List[ProjectTypeStats]:
```
Hagellach37, line 123: I'm not sure if this function name is clear. As I understand it you are running a query for the build_area project type here not for all project types. I would suggest to change the name.

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+        )
+    return project_list
+
+
+def get_project_swipe_type() -> List[ProjectSwipeTypeStats]:
+    PROJECT_TYPE_STATS_QUERY = f"""
+        SELECT
+            P.project_type as project_type,
+            COUNT(*) as total_swipe
+            FROM {Result._meta.db_table} R
```
Hagellach37, line 149: I'm not sure if you can use a simple count here as the app interface is different for the project types. Here an overview how the number of results relates to the number of swipes:

- project type 1 and 4: 6 tasks --> 1 swipe
- project type 2 and 3: 1 task --> 1 swipe

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+    for project_type, total_swipe in aggregate_results:
+        project_list.append(
+            ProjectSwipeTypeStats(
+                total_swipe=total_swipe or 0, project_type=project_type
+            )
+        )
+    return project_list
+
+
+def get_project_geo_area_type() -> List[MapContributionTypeStats]:
```
Hagellach37, line 166: could you clarify what `project_geo_area_type` refers to?

django/apps/existing_database/utils.py
```diff
@@ -0,0 +1,14 @@
+def parse_geom(geom):
+    # NOTE: This method transform data equivalent to what mapbox supports
+    # Since data structure from postgres isn't mapbox compatible
```
Hagellach37, line 3: not sure if we still need this. Do we use mapbox at some other part still?

firebase/database.rules.json
```diff
@@ -133,6 +133,9 @@
             ".write": "auth != null && auth.uid == $uid"
           }
         },
+        "userGroups": {
+          ".write": "auth != null"
+        },
```
Hagellach37, line 138: Is this needed? It looks like you define it again a few lines below.

firebase/functions/src/osm_auth.ts
```diff
@@ -10,27 +10,26 @@
...
+import * as crypto from 'crypto';
+import * as cookieParser from 'cookie-parser';
+import * as simpleOAuth2 from 'simple-oauth2';
+import axios from 'axios';
 
 // this redirect_uri MUST match the one set in the app on OSM OAuth, or you
 // will get a cryptic error about the server not being able to continue
 // TODO: adjust the prefix based on which deployment is done (prod/dev)
-const OAUTH_REDIRECT_URI = functions.config().osm.redirect_uri;
+const OAUTH_REDIRECT_URI = functions.config().osm?.redirect_uri;
```
Hagellach37, line 22: just out of curiosity: what does the `?` mean here. why did you add it?

manager-dashboard/.env.example
```diff
@@ -0,0 +1,10 @@
+# REACT_APP_SENTRY_DSN=
+# REACT_APP_ENVIRONMENT=
+
+REACT_APP_FIREBASE_API_KEY=
+REACT_APP_FIREBASE_AUTH_DOMAIN=
+REACT_APP_FIREBASE_DATABASE_URL=
+REACT_APP_FIREBASE_PROJECT_ID=
+REACT_APP_FIREBASE_STORAGE_BUCKET=
+REACT_APP_FIREBASE_MESSAGING_SENDER_ID=
+REACT_APP_FIREBASE_APP_ID=
```
Hagellach37, line 10: here you define another .env file. Would it make sense to use the existing one or was there any particular reason to have it like this?

manager-dashboard/.github/pull_request_template.md
```diff
@@ -0,0 +1,20 @@
+- Addresses XXX
```
Hagellach37, line 1: a general question: would we use the existing `.github/` folder in the repo instead of this one?

manager-dashboard/.github/workflows/ci.yml
```diff
@@ -0,0 +1,72 @@
+name: CI
```
Hagellach37, line 1: same question as above: where would be the best place for this ci.yml?

Hagellach37 (commented):
Hey,

I've left a first round of comments. It would be great if you can take a look at these and help me to understand properly what you did. :)

I covered the django and firebase parts. For the other parts I will add more comments soon,

manager-dashboard/app/components/TileServerInput/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+import TextInput from '#components/TextInput';
+import RadioInput from '#components/RadioInput';
+import {
+    valueSelector,
+    labelSelector,
+    getNoMoreThanNCharacterCondition,
+} from '#utils/common';
+
+export type TileServerType = 'bing' | 'mapbox' | 'maxar_standard' | 'maxar_premium' | 'esri' | 'esri_beta' | 'sinergise' | 'custom';
```
Hagellach37, line 21: we can remove all references to `singergise`. imagery from Sinergise is not provided anymore

manager-dashboard/app/components/TutorialList/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+export type ProjectType = 1 | 2 | 3 | 4;
+const projectTypeLabelMap: {
+    [key in ProjectType]: string
+} = {
+    1: 'Build Area',
+    2: 'Footprint',
+    3: 'Change Detection',
+    4: 'Completeness',
+};
+
```
Hagellach37, line 29: yes. I would encourage to define this somewhere else as these might get used in other parts of the manager dashboard as well.

manager-dashboard/app/components/GeoJsonFileInput/index.tsx
```diff
@@ -0,0 +1,159 @@
+import React from 'react';
+import booleanValid from '@turf/boolean-valid';
+
+import FileInput, { Props as FileInputProps } from '#components/FileInput';
+import useMountedRef from '#hooks/useMountedRef';
+import GeoJsonPreview from '#components/GeoJsonPreview';
+
+function validateGeoJSON(value: GeoJSON.GeoJSON): boolean {
```
Hagellach37, line 8: In the existing manager dashboard we check how many features the geojson file contains and also check for the size of the area of interest depending on the zoom level defined by the project manager (not only the size of the file). Maybe I've missed this part, could you clarify if you implemented these checks as well?

manager-dashboard/app/views/Login/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    finalValues.email as string,
+                    finalValues.password as string,
+                );
+                // NOTE: we will udpate the current user on <Init />
+                if (!mountedRef.current) {
+                    return;
+                }
+                setErrorMessage(undefined);
+                setPending(false);
+            } catch (submissionError) {
```
Hagellach37, line 88: just so that I can better understand this part: Where do you define that only project managers can login through the manager dashboard? Normal mapswipe users will not be able to see/login for the manager dashboard, only user account with project manager credentials in firebase will be able to do this. Is this handled here or in another file?

Hagellach37 (commented):
just adding another round of comments. will provide more soon

mapswipe-dashboard/package.json
```diff
@@ -0,0 +1,137 @@
+{
+    "name": "mapswipe-manager-dashboard",
+    "longName": "MapSwipe Manager Dashboard",
+    "version": "1.0.0",
+    "description": "Manager Dashboard for MapSwipe",
```
Hagellach37, line 5: this shouldn't be `manager-dashboard`. also check the other references below and the git repo url etc.

mapswipe_workers/mapswipe_workers/auth.py
```diff
@@ -47,17 +47,20 @@ def firebaseDB() -> object:
 class postgresDB(object):
     """Helper class for Postgres interactions"""
 
-    _db_connection = None
+    __db_connection = None
```
Hagellach37, line 50: it's not really clear to me why this has changed here.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -340,6 +384,79 @@ def save_results_to_postgres(results_file, project_id, filter_mode: bool):
...
+                AND R1.user_id = R2.user_id
+        """
+        p_con.query(filter_query, {"project_id": project_id})
+
+    query_insert_results = """
+        INSERT INTO results_user_groups
+            SELECT * FROM results_user_groups_temp
+        ON CONFLICT (project_id, group_id, user_id, user_group_id)
+        DO NOTHING;
+        -- TRUNCATE results_user_groups_temp; ??
```
Hagellach37, line 453: this is not needed, as we run this in a separate step.

nginx/nginx.conf.template
```diff
@@ -0,0 +1,70 @@
...
+        proxy_pass  http://manager_dashboard:80/;
+    }
+
+    location /manager_dashboard {
+        rewrite ^ /manager_dashboard/ permanent;
+    }
+
+    location / {
+        rewrite ^ /manager_dashboard/ permanent;
+    }
```
Hagellach37, line 38: this needs to be updated and should use the new manager dashboard and mapswipe dashboard

Hagellach37 (commented):
hey, and here the final round of comments.

Thanks for all this work. :) It's really impressive and I think you basically touched all parts of MapSwipe.

firebase/functions/src/osm_auth.ts
laurentS: It checks for undefined/null values, so the above checks that `osm` is not null/undefined before trying to read `redirect_uri` from it, otherwise, it defaults to undefined instead of throwing an error.
I think this is ok if there's a check for undefined result, as the code should not run without these values set.

mapswipe-dashboard/BUILD.md
```diff
@@ -0,0 +1,43 @@
+## Webpack
```
laurentS, line 1: what do these mean? Can you add a bit of explanation at the start of the file to clarify?

mapswipe-dashboard/app/Base/configs/googleAnalytics.ts
```diff
@@ -0,0 +1,12 @@
+import { InitializeOptions } from 'react-ga';
```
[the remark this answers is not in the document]
laurentS: I would remove GA. We don't need google to track our users more than they already do. And GA has also been deemed illegal under GDPR by various EU oversight bodies.

mapswipe-dashboard/app/components/Header/index.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { memo } from 'react';
+import { _cs } from '@togglecorp/fujs';
```
laurentS, line 2: Is this package meant to stay or is it an internal one for development only? There is no README on your github repo, so it's very hard to know what this does.

mapswipe_workers/Dockerfile
```diff
@@ -1,3 +1,4 @@
+
```
laurentS, line 1: blank line not needed.

mapswipe_workers/mapswipe_workers/auth.py
```diff
@@ -47,17 +47,20 @@ def firebaseDB() -> object:
...
-        self._db_connection = psycopg2.connect(
-            database=POSTGRES_DB,
-            host=POSTGRES_HOST,
-            password=POSTGRES_PASSWORD,
-            port=POSTGRES_PORT,
-            user=POSTGRES_USER,
-        )
+    @property
+    def _db_connection(self):
+        if self.__db_connection is None:
```
laurentS, line 55: This will setup the connection when first called, as opposed to startup time. I'm wondering if startup time isn't better in the sense that it will raise an error immediately (ie. when the admin is starting the app and paying attention), rather than at some random point later in time, when maybe they're gone and doing something else?
Any particular reason to do it this way?

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -82,15 +82,30 @@ def transfer_results_for_project(project_id, results, filter_mode: bool = False)
         # The user_id is used as a key in the postgres database for the results
         # and thus users need to be inserted before results get inserted.
         results_user_id_list = get_user_ids_from_results(results)
+        results_user_group_id_list = set(
```
laurentS, line 85: Can you add a return type annotation to the functions you're modifying, so that `mypy` will check them?

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
```diff
@@ -221,12 +240,20 @@ def results_to_file(results, projectId):
...
             end_time = dateutil.parser.parse(end_time)
             timestamp = end_time
+            results_added = False
 
-            if type(results) is dict:
-                for taskId, result in results.items():
+            if type(result_results) is dict:
+                if result_results:
+                    results_added = True
+                for taskId, result in result_results.items():
```
laurentS, line 256: You're reusing `result` as a variable name here, this will cause problems.

mapswipe_workers/requirements.txt
```diff
@@ -4,12 +4,12 @@ click==8.1.3
 firebase-admin==4.4.0
 flake8==3.8.3
 mapswipe-workers==3.0
-pandas==1.4.3
+pandas==1.1.5
```
laurentS, line 7: Why are you downgrading `pandas`?

mapswipe_workers/setup.cfg
```diff
@@ -9,7 +9,7 @@
 # These settings are compatible with Black
 ignore = W503, E203
 inline-quotes = double
-max-line-length = 88
+max-line-length = 119
```
laurentS, line 12: Any reason why you're changing this?

mapswipe_workers/tests/unittests/fixtures/tile_map_service_grid/users/users_-M7nTGflMusNxFd4iSfD.csv
```diff
@@ -1,247 +1,247 @@
...
-243,-M7nTGflMusNxFd4iSfD,zs9IdLDJ0hfnWAHc9PF1NiMnX3k1,1,291,820,53,0.9392898052691867
-244,-M7nTGflMusNxFd4iSfD,ztgbNPjRjTUuOKlknwDgejvpxXo2,1,291,580,2,0.9965635738831615
-245,-M7nTGflMusNxFd4iSfD,zxS3YtVzToRSmBsKT6k5P2t837I2,3,867,1393,341,0.8033448673587081
+idx,project_id,user_id,username,groups_completed,total_contributions,agreeing_contributions,disagreeing_contributions,simple_agreement_score
+0,-M7nTGflMusNxFd4iSfD,08tsiIBXUNOBYtO1ZwmGU9t1iKg1,Test-user,7,2025,2769,1569,0.6383125864453665
+1,-M7nTGflMusNxFd4iSfD,0rluq7FANKRXsiNFNDRriq5PVZ72,Test-user,2,579,1151,7,0.9939550949913644
+2,-M7nTGflMusNxFd4iSfD,0sbtU3vn1rSvD7dY3zIbOO19bWy2,Test-user,3,867,1376,649,0.6795061728395062
+3,-M7nTGflMusNxFd4iSfD,13MrH4g49oRR1oyAjl0DfVw9KBQ2,Test-user,5,1452,2026,878,0.6976584022038568
+4,-M7nTGflMusNxFd4iSfD,1E0qG4OyDiXyhClCyQtQvCTpypD2,Test-user,23,6648,10834,3041,0.7808288288288289
+5,-M7nTGflMusNxFd4iSfD,1O6U2O6LTuhMHROAtBG9EcBqrJQ2,Test-user,1,288,788,76,0.9120370370370372
```
laurentS, line 7: This result is changed at the 16th decimal. I'm assuming other results change in a similar way. Is it important? Why does the result change? @Hagellach37 what do you think?

laurentS: I haven't checked the previous CSV file, as the diff is almost a million lines long :(

laurentS (changes requested):
Are you planning on adding tests to the django module? Even a few simple requests are useful to avoid introducing major bugs.
Also, can you add `mypy`? I ran it locally on the django code and it listed 120+ errors with the default options.

laurentS:
To help with maintenance/deployment, I've found that using a `Makefile` with the various useful commands is quite nice:
- at the prompt, you just type `make <tab>` and you get a list of all the useful commands needed to manage the project,
- if you give long descriptive names, they're self explanatory, and you can add comments in the `Makefile` to expand further,
- because of auto-completion in bash, it's still easy to type,
- it works pretty much anywhere, no dependencies needed.

django/apps/existing_database/management/commands/update_organization_name.py
```diff
@@ -0,0 +1,37 @@
...
+    def handle(self, *args, **kwargs):
+        with connections[settings.MAPSWIPE_EXISTING_DB].cursor() as cursor:
+            cursor.execute(self.FETCH_ORGANIZATION_DETAILS)
+            aggregate_results = cursor.fetchall()
+        for project_id, name in aggregate_results:
+            organization_name = self.parse_organization_name(name)
+            with connections[settings.MAPSWIPE_EXISTING_DB].cursor() as cursor:
+                cursor.execute(
+                    f"""
+                        UPDATE projects set organization_name='{organization_name}'
```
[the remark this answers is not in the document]
thenav56: Makes sense. We have replaced the SQL with Django ORM. So this should be fixed.
https://github.com/mapswipe/python-mapswipe-workers/blob/0d013617a60b747643ae21bd9b42fbff295fd687/django/apps/existing_database/management/commands/update_organization_name.py#L18-L36

django/apps/existing_database/models.py
thenav56: I will recommend using JSONB. But maybe we can discuss this later.

> And this has some immediate benefits:
> - More efficiency.
> - Significantly faster to process.
> - Supports indexing (which can be a significant advantage, as we’ll see later). simpler schema designs (replacing entity-attribute-value (EAV) tables with JSONB columns, which can be queried, indexed, and joined, allowing for performance improvements up until 1000X!)
> 
> Few Drawbacks:
> - Slightly slower input (due to added conversion overhead).
> - It may take more disk space than plain JSON due to a larger table footprint, though not always.

django/apps/existing_database/models.py
thenav56: A mistake on our side. It's fixed.
https://github.com/mapswipe/python-mapswipe-workers/blob/0d013617a60b747643ae21bd9b42fbff295fd687/django/apps/existing_database/models.py#L56-L91

django/apps/existing_database/models.py
thenav56: Django needs to have a single primary key for maintaining relations. Due to multiple keys as primary in task and group, it isn't possible for now.

django/apps/existing_database/query.py
thenav56: This has been changed. Can you maybe review this new file where we do this aggregation?
https://github.com/mapswipe/python-mapswipe-workers/blob/project/group-analytics/django/apps/aggregated/management/commands/update_aggregated_data.py

django/apps/existing_database/query.py
thenav56: There was a lot of inconsistency in the naming. We have changed a lot. Can you re-check? Sorry about this.

django/apps/existing_database/query.py
thenav56: This has been changed. Can you maybe review this new file where we do this aggregation?
https://github.com/mapswipe/python-mapswipe-workers/blob/0d013617a60b747643ae21bd9b42fbff295fd687/django/apps/aggregated/management/commands/update_aggregated_data.py#L74-L83

**I will add project type 4 there as well.**
```sql
WHEN P.project_type in (1, 4) THEN ROUND(task_count/6)
```

django/apps/existing_database/utils.py
thenav56: We have removed this.

manager-dashboard/.env.example
thenav56: These are for specific components. Maybe we can discuss further on this.

manager-dashboard/.github/pull_request_template.md
thenav56: These are for specific components (pulled when we merged another repo to this one). Maybe we can discuss further on this.

manager-dashboard/.github/workflows/ci.yml
thenav56: These are for specific components (pulled when we merged another repo to this one). Maybe we can discuss further on this.

mapswipe_workers/mapswipe_workers/auth.py
thenav56: This was done for fixing pyright issues. It's an optional change.
image

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
thenav56: Sure. Will remove this.

nginx/nginx.conf.template
thenav56: This has been changed. Can you re-check this new one?
https://github.com/mapswipe/python-mapswipe-workers/blob/0d013617a60b747643ae21bd9b42fbff295fd687/nginx/nginx.conf.template#L11-L27

.github/workflows/actions.yml
```diff
@@ -24,9 +24,9 @@ jobs:
     - name: Code style
       working-directory: ./mapswipe_workers
```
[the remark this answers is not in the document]
thenav56: Let's talk about this further.

For now, the Django directory is also being checked.

.github/workflows/actions.yml
```diff
@@ -60,5 +60,5 @@ jobs:
         POSTGRES_DB: test
         OSMCHA_API_KEY: ${{ secrets.OSMCHA_API_KEY }}
       run: |
-        docker-compose run mapswipe_workers_creation python -m unittest discover --verbose --start-directory tests/unittests/
-        docker-compose run mapswipe_workers_creation python -m unittest discover --verbose --start-directory tests/integration/
\ No newline at end of file
+        docker-compose run --rm mapswipe_workers_creation python -m unittest discover --verbose --start-directory tests/unittests/
+        docker-compose run --rm mapswipe_workers_creation bash -c 'pip install pytest && py.test -ra -v --durations=10 tests/integration/'
```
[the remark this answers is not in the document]
thenav56: I missed the change notice. Will use pytest.

I think using the latest pytest shouldn't have any impact as there is not much here in the code related to it. But if you think otherwise let us know.
Let's discuss further about dependency management for mapswipe_workers.

django/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+COPY pyproject.toml poetry.lock /code/
+
+RUN apt-get update -y \
+    && apt-get install -y \
+        # Used for waiting for postgres db.
+        wait-for-it \
+        # For postgis
+        gdal-bin \
+    # Upgrade pip and install python packages for code
+    && pip install --upgrade --no-cache-dir pip poetry \
```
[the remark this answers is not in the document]
thenav56: Sure. Will add the version there. We are using poetry version 1.2.1.

django/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+        && poetry --version \
+        # Configure to use system instead of virtualenvs
+        && poetry config virtualenvs.create false \
+        && poetry install --no-root \
+        # Remove installer
+        && pip uninstall -y poetry virtualenv-clone virtualenv \
```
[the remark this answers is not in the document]
thenav56: Clean-up is just something we have there as a standard procedure.

For the use of -buster, we got into an issue with installing psycopg2 with slim so we are using the normal one for now.

django/README.md
```diff
@@ -0,0 +1,29 @@
+## Host (Optional)
```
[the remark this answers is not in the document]
thenav56: Let's discuss this further.

django/apps/existing_database/dataloaders.py
```diff
@@ -0,0 +1,931 @@
...
+        user_id,
+        user_group_id,
+        user_group_name,
+        COUNT(user_id) as members_count
+    FROM user_data
+    GROUP BY user_id, user_group
+"""
+
+
+def load_user_group_stats(keys: List[str]):
```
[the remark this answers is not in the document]
thenav56: Sure.

django/apps/existing_database/dataloaders.py
```diff
@@ -0,0 +1,931 @@
...
+                total_swipes=total_swipe_count or 0,
+                total_swipe_time=round(total_swipe_count / 60)
+                or 0,  # swipe time in minutes
+                total_mapping_projects=mapped_project_count or 0,
+            )
+        )
+    return [_map.get(key) for key in keys]
+
+
+# def load_user_group_user_contributors_stats(keys: List[str]):
```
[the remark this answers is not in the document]
thenav56: Not needed. We have removed it.

django/apps/existing_database/dataloaders.py
```diff
@@ -0,0 +1,931 @@
...
+    return [_map.get(key) for key in keys]
+
+
+def load_user_group_user_stats(keys: List[str]):
+    """
+    Load user stats under user_group
+    """
+    with connections[settings.MAPSWIPE_EXISTING_DB].cursor() as cursor:
+        cursor.execute(USER_GROUP_USER_SWIPE_STAT_QUERY, [keys])
+        aggregate_results = cursor.fetchall()
```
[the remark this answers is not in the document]
thenav56: We have replaced most of the raw SQL with Django ORM.

django/apps/existing_database/dataloaders.py
```diff
@@ -0,0 +1,931 @@
...
+
+    @cached_property
+    def load_user_group_contributors_stats(self):
+        return DataLoader(load_fn=sync_to_async(load_user_group_contributors_stats))
+
+    @cached_property
+    def load_user_group_project_type_stats(self):
+        return DataLoader(load_fn=sync_to_async(load_user_group_project_type_stats))
+
+    # @cached_property
```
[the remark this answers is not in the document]
thenav56: We have removed it.

django/apps/existing_database/factories.py
```diff
@@ -0,0 +1,105 @@
...
+class TaskFactory(DjangoModelFactory):
+    task_id = factory.Sequence(lambda n: f"dummy-project-task-{n}")
+
+    class Meta:
+        model = Task
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        if "group" not in kwargs:
+            raise Exception("Please define group")
```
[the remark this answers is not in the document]
thenav56: This part is mostly for showing errors if the factories are not used properly by a dev. Not designed to be caught on calling code. But if you insist then let me know.

django/apps/existing_database/management/commands/update_organization_name.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def parse_organization_name(self, name):
+        if name and "\n" in name:
+            name_parts = [x.strip() for x in name.split("\n")]
+            if name_parts[1] == "":
+                return "null"
```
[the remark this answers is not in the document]
thenav56: We have replaced this with an empty string.

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+            COUNT( DISTINCT user_group_id) as total_groups,
+            COUNT(DISTINCT user_id) as total_users,
+            SUM(swipe_count) as total_swipes
+        From dashboard_data
+    """
+    with connections[settings.MAPSWIPE_EXISTING_DB].cursor() as cursor:
+        cursor.execute(COMMUNITY_DAHBOARD_STATS_QUERY)
+        aggregate_results = cursor.fetchall()
+    for data in aggregate_results:
+        return CommunityStatsType(
```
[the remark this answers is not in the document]
thenav56: This has been fixed.

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+            COUNT( DISTINCT user_group_id) as total_groups,
+            COUNT(DISTINCT user_id) as total_users,
+            SUM(swipe_count) as total_swipes
+        From dashboard_data
+    """
+    with connections[settings.MAPSWIPE_EXISTING_DB].cursor() as cursor:
+        cursor.execute(COMMUNITY_DAHBOARD_STATS_QUERY)
+        aggregate_results = cursor.fetchall()
+    for data in aggregate_results:
+        return CommunityStatsLatestType(
```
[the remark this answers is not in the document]
thenav56: This has been fixed.

django/apps/existing_database/query.py
```diff
@@ -0,0 +1,259 @@
...
+            WHERE P.geom is not null AND P.project_type = '1'
+            GROUP BY P.project_type
+            """
+    with connections[settings.MAPSWIPE_EXISTING_DB].cursor() as cursor:
+        cursor.execute(PROJECT_TYPE_STATS_QUERY)
+        aggregate_results = cursor.fetchall()
+    project_list = []
+    for project_type, area_sum in aggregate_results:
+        project_list.append(
+            ProjectTypeStats(area=area_sum / 1000000, project_type=project_type)
```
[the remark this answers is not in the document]
thenav56: This has been removed.

django/apps/existing_database/types.py
```diff
@@ -0,0 +1,332 @@
...
+    @strawberry.field
+    async def stats(self, info: Info, root: User) -> Optional[UserSwipeStatType]:
+        return await info.context["dl"].existing_database.load_user_stats.load(
+            root.user_id
+        )
+
+    @strawberry.field
+    async def contribution_stats(
+        self, info: Info, root: User
+    ) -> Optional[List[ContributorType]]:
```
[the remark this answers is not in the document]
thenav56: Sure.

django/pyproject.toml
```diff
@@ -0,0 +1,30 @@
...
+Django = "^4.0.5"
+django-environ = "*"
+django-cors-headers = "*"
+psycopg2 = "^2.9.3"
+factory-boy = "^3.2.1"
+gunicorn = "^20.1.0"
+strawberry-graphql-django = "==0.3"
+strawberry-graphql = "^0.114.1"
+strawberry-django-plus = "^1.14.1"
+starlette = "^0.20.3"
```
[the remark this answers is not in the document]
thenav56: Looks like is it used in the code, but for typing only. Removing it now.

django/apps/existing_database/factories.py
```diff
@@ -0,0 +1,105 @@
...
+    Task,
+    User,
+    UserGroup,
+    UserGroupResult,
+    UserGroupUserMembership,
+)
+
+
+class UserFactory(DjangoModelFactory):
+    user_id = factory.Sequence(lambda n: f"dummy-user-id-{n}")
```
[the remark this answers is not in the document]
thenav56: This sounds interesting. Let's discuss this further later.

django/scripts/run_develop.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -x
+
+python3 manage.py runserver 0.0.0.0:80
```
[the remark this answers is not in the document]
thenav56: We don't expose this to the host in production. For development, we are using 8113 through docker.
I will change this to 8113 here as well.

docker-compose.tc.yaml
```diff
@@ -0,0 +1,128 @@
...
+    IMAGE_MAXAR_STANDARD_API_KEY: '${IMAGE_MAXAR_STANDARD_API_KEY}'
+    SLACK_TOKEN: '${SLACK_TOKEN}'
+    SLACK_CHANNEL: '${SLACK_CHANNEL}'
+    SENTRY_DSN: '${SENTRY_DSN}'
+    OSMCHA_API_KEY: '${OSMCHA_API_KEY}'
+  depends_on:
+    - postgres
+  volumes:
+    - ./mapswipe_workers/mapswipe_workers:/usr/local/lib/python3.10/dist-packages/mapswipe_workers/:ro
+    - ./mapswipe_workers/mapswipe_workers:/usr/local/lib/python3.8/dist-packages/mapswipe_workers/:ro
```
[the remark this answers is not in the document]
thenav56: For development, we don't want to build it every time it was changed.
Also, the build does pip after `COPY . .` which takes a long time to run.
So for now just mapping the host code to the container for local development.

mapswipe_workers/Dockerfile
thenav56: Sure.

mapswipe_workers/mapswipe_workers/auth.py
thenav56: Also, the reason was for defining _db_connectin as a property (for using _db_connectin as an object field instead of a class field). This is used in a bit hacky way to resolve DB connection open/close issues. Maybe we can discuss this further for a better test framework.

mapswipe_workers/requirements.txt
thenav56: worker docker build was failing due to pandas==1.4.3. Which was solved by downgrading pandas.

mapswipe_workers/setup.cfg
thenav56: Seems more natural to use around 120 due to our current screen sizes. :D

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
thenav56: Good catch.

mapswipe_workers/mapswipe_workers/auth.py
thenav56: https://github.com/mapswipe/python-mapswipe-workers/pull/562#discussion_r992123210

I think existing uses also create a similar problem as the function is called in each module. Maybe we can discuss it further on this issue.

mapswipe_workers/mapswipe_workers/firebase_to_postgres/transfer_results.py
thenav56: Not sure what you meant here.

mapswipe_workers/tests/unittests/fixtures/tile_map_service_grid/users/users_-M7nTGflMusNxFd4iSfD.csv
thenav56: username field was added there. Not sure why we have 448597 rows here.

django/apps/existing_database/models.py
tnagorra: @Hagellach37 should we create a ticket for these kind of improvements?

manager-dashboard/app/components/GeoJsonFileInput/index.tsx
tnagorra: Yes, we have implemented the check.
We have a validation function defined here https://github.com/mapswipe/python-mapswipe-workers/blob/0d013617a60b747643ae21bd9b42fbff295fd687/manager-dashboard/app/views/NewProject/utils.ts#L104-L167

manager-dashboard/app/views/Login/index.tsx
tnagorra: It's defined on a separate file.

https://github.com/mapswipe/python-mapswipe-workers/blob/0d013617a60b747643ae21bd9b42fbff295fd687/manager-dashboard/app/Base/components/Init/index.tsx#L67-L77

manager-dashboard/app/views/NewProject/utils.ts
```diff
@@ -0,0 +1,443 @@
...
+>;
+
+type ProjectFormSchema = ObjectSchema<PartialProjectFormType>;
+type ProjectFormSchemaFields = ReturnType<ProjectFormSchema['fields']>;
+
+// FIXME: break this into multiple geometry conditions
+const DEFAULT_MAX_FEATURES = 20;
+// const DEFAULT_MAX_FEATURES = 10;
+const DEFAULT_MAX_AREA = 20;
+function validGeometryCondition(
```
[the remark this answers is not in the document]
tnagorra: It does indeed! :rocket:

django/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+# Install the used python version if not already
+pyenv install $(cat .python-version)
+
+poetry install
+# For psycopg2, install dependencies (For arch: https://github.com/archlinux/svntogit-community/blob/packages/python-psycopg2/trunk/PKGBUILD)
+```
+
+
+### Add new app in django
```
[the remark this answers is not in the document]
tnagorra: Let's discuss further on this.

mapswipe-dashboard/BUILD.md
tnagorra: We will remove this. This is just a list of dependencies used on the initial template.

mapswipe-dashboard/app/components/Header/index.tsx
tnagorra: This is meant to stay. We do have a documentation but it's hard to find.
https://github.com/toggle-corp/fujs/blob/develop/mydocs/modules.md

firebase/database.rules.json
frozenhelium: We need this, because one is for overall usergroup and one is for the membership

Issue #656, opened 2023-02-27
delete projects is not working due to foreign key constraint


Hagellach37:
introducing the new tables for aggregated stats, breaks our workflow to delete projects.

```
Traceback (most recent call last):
  File "/usr/local/bin/mapswipe_workers", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.8/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/mapswipe_workers/mapswipe_workers.py", line 460, in run_delete_project
    if delete_project.delete_project(project_ids):
  File "/usr/local/lib/python3.8/dist-packages/mapswipe_workers/firebase_to_postgres/delete_project.py", line 115, in delete_project
    pg_db.query(sql_query, {"project_id": project_id})
  File "/usr/local/lib/python3.8/dist-packages/mapswipe_workers/auth.py", line 68, in query
    self._db_connection.commit()
psycopg2.errors.ForeignKeyViolation: update or delete on table "projects" violates foreign key constraint "aggregated_aggregate_project_id_01e6f153_fk_projects_" on table "aggregated_aggregateduserstatdata"
DETAIL:  Key (project_id)=(-NOJYMJyWFMwZtxqW0Ms) is still referenced from table "aggregated_aggregateduserstatdata".
```

PR #657, opened 2023-02-28
Feature/fix project delete issue

thenav56:
Address
- https://github.com/mapswipe/python-mapswipe-workers/issues/656

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -36,7 +36,7 @@ x-django: &base_django
       DJANGO_MEDIA_ROOT: '/assets/media/'
       SENTRY_DSN: ${DJANGO_SENTRY_DSN}
       SENTRY_SAMPLE_RATE: ${DJANGO_SENTRY_SAMPLE_RATE:-0.2}
-      MAPSWIPE_ENVIRONMENT: ${MAPSWIPE_ENVIRONMENT:-dev}
+      MAPSWIPE_ENVIRONMENT: ${MAPSWIPE_ENVIRONMENT:-prod}
       RELEASE: '${DJANGO_RELEASE:-latest}'
   volumes:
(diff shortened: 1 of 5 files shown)


docker-compose.yaml
```diff
@@ -36,7 +36,7 @@ x-django: &base_django
       DJANGO_MEDIA_ROOT: '/assets/media/'
       SENTRY_DSN: ${DJANGO_SENTRY_DSN}
       SENTRY_SAMPLE_RATE: ${DJANGO_SENTRY_SAMPLE_RATE:-0.2}
-      MAPSWIPE_ENVIRONMENT: ${MAPSWIPE_ENVIRONMENT:-dev}
+      MAPSWIPE_ENVIRONMENT: ${MAPSWIPE_ENVIRONMENT:-prod}
```
laurentS, line 39: Why the change to prod here? I would have thought that keeping things to dev by default would make it easier to start a dev env for newcomers, and also avoid any accidental actions on prod for users who happen to have prod info available.

laurentS (approved):
Looks ok to me. I'm baffled by the fact that django doesn't support on delete constraints!

docker-compose.yaml
thenav56: Currently **docker-compose.yaml** is not development friendly as we need to build for each change and also the nginx configs require some custom work. That is why we created **docker-compose.tc.yaml** to fit our development need which is also WIP.
So, as we are using **docker-compose.yaml** in the production run, I used the environment to be _prod_ by default. We should define this accordingly in staging/production. But also we are using separate sentry projects in staging/production, this is not really used.

thenav56: > I'm baffled by the fact that django doesn't support on delete constraints!

Same here.

PR #662, opened 2023-03-02
Use a different geojson validator

tnagorra:
We are currently using `@turf/boolean-valid` for validation of geojson.
We have a false negative for a geojson. (For more information https://mapswipe.slack.com/archives/C0HCW2CJJ/p1677586146742749?thread_ts=1677518532.043429&cid=C0HCW2CJJ)

The library does not specify what is wrong with the geojson which is not very useful.

Changing the geojson validation library from `@turf/boolean-valid` to `@placemarkio/check-geojson` so that we get better errors and avoid issue with the false negative.

### DIFF
--- a/manager-dashboard/app/components/GeoJsonFileInput/index.tsx
+++ b/manager-dashboard/app/components/GeoJsonFileInput/index.tsx
@@ -1,32 +1,31 @@
 import React from 'react';
-import booleanValid from '@turf/boolean-valid';
+import { check, HintError } from '@placemarkio/check-geojson';
 
 import FileInput, { Props as FileInputProps } from '#components/FileInput';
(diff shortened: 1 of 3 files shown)


frozenhelium (approved):

frozenhelium (approved):

frozenhelium:
@tnagorra Should we merge this?

danbjoseph:
would also be useful if you could note the reason for the change in geojson validator - i.e. which issue(s) is this solving?

tnagorra:
The conversation is on slack at https://mapswipe.slack.com/archives/C0HCW2CJJ/p1677643592304189?thread_ts=1677518532.043429&cid=C0HCW2CJJ

I have also updated the description of the PR

PR #667, opened 2023-03-20
Add script for backtrack calculation

thenav56:
Issue:
- https://github.com/mapswipe/mapswipe-meta/issues/13

## NOTE
Before running the SQL script, generate a user-group mapping session table CSV dump (If we need to revert back changes)
```bash
docker-compose exec postgres psql -U mapswipe_workers mapswipe -c '\copy (SELECT * FROM mapping_sessions_user_groups) to STDOUT CSV HEADER;' > /tmp/user-group-session-data-`date +%Y-%m-%d.%H:%M:%S`.csv
```
Running the SQL Script and aggregated data calculation
```bash
docker-compose exec -T postgres bash -c 'psql -U $POSTGRES_USER $POSTGRES_DB' < postgres/scripts/backtrack_user_group_mapping_sessions.sql
docker-compose exec django ./manage.py update_aggregated_data
```

### DIFF
--- a/postgres/scripts/backtrack_user_group_mapping_sessions.sql
+++ b/postgres/scripts/backtrack_user_group_mapping_sessions.sql
@@ -1,0 +1,64 @@
+-- https://docs.google.com/spreadsheets/d/1g92Rj02LVqvYBFmdnf5CBs8IwuD0GjXP0b1A-H9SvQY/edit#gid=0
+
+WITH
+    -- Collect user_group_id
+    cq_user_group_ids as (
+        SELECT
+            user_groups.user_group_id
+        FROM user_groups
+        WHERE user_groups.name in (
+            'Amazon',
+            'Cisco',
+            'Grainger',
+            'Hewlett Packard Enterprise',
+            'HP Inc.',
+            'Intel',
+            'Lenovo',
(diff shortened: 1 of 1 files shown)


danbjoseph:
@thenav56, you've requested Benni's review but the PR is still marked as a draft. can you please update the PR to match the current state and add any information that will be useful for Benni. for example, what needs to happen after this is merged? thank you.

thenav56:
Hi @danbjoseph
The reason behind the PR being drafted was that we thought you would be interested in a preview before deploying this to prod as this will include data migrations which will need additional work if we need to revert back or change something. This was included in the [bi-weekly meetings](https://docs.google.com/document/d/1-cI6F0qZ2ikD62N1SMiZXMXPEs-PTMvgVrX9e1DThFE/edit?disco=AAAAtIMm-HY).

If no preview is required, we can go ahead with the review/merge and deployment. I have added additional steps in PR description to produce a CSV snapshot of the table before running the migration if we need to revert back.

NOTE: By preview, we meant to run the migration in our internal-hosted dashboard using the latest database dump and share it with you using https://mapswipe-community-dashboard.dev.togglecorp.com/

PR #780, opened 2023-05-12
Pre-calculate total geo area and time max limit for project groups

thenav56:
Deployment steps
- Run SQL migration script
```
postgres/scripts/v2_to_v3/06_alter_groups_data_add_total_area_and_time_limit_columns.sql
```
- Run aggregated data collection
```bash
docker-compose run --rm django ./manage.py update_aggregated_data.py
```


mapswipe_workers/mapswipe_workers/project_types/base/project.py
```diff
@@ -386,7 +386,15 @@ def save_to_postgres(self, project, groups, groupsOfTasks):
...
-            INSERT INTO groups
+            INSERT INTO groups (
+              project_id,
+              group_id,
+              number_of_tasks,
+              finished_count,
+              required_count,
+              progress,
+              project_type_specifics
+            )
```
Hagellach37, line 397: I would have expected that we add the total group area max time here as well? Or is this done in another step?

thenav56: We are using Django for calculating this part. I have also added a comment mentioning that as well.
This can take some time to calculate and also doesn't work with existing ones as well.
So for now doing this on Django made more sense as it is calculated and maintained by Aggregated logic.

Hagellach37: mhh okay, not sure if I prefer this way to split it up. Could you investigate how long this will take for a single project during project creation.

I think it's "just" adjusting the sql statement slightly and summing up the task area per group with postgis. @ElJocho might be able to advise here as well.

Hagellach37: ideally the group information could be set directly when setting up the project, then we do not need to rely on django to perform additional steps at some point in time.

for backtrack calculation of existing projects your approach is good. I'm just concerned about new projects. 

It is already difficult to understand which information is set for which project, e.g. considering the various project types. @ElJocho has worked towards refactoring this part of the code so that it will be easier to learn which attributes are expected.

thenav56: That makes sense, I wanted to add this to the group as well. Considering the refactor and complex structure right now I wanted to make minimal changes to the mapswipe_workers core part as well. Currently, the calculation/test parts also fit perfectly in the aggregated module as it is tightly coupled with it and only used there.

Maybe @ElJocho can move this aggregated logic to the core part when during the refactor. The SQL query is already defined, will just need to add it to the core logic and maybe define further tests as well. Which seems to be perfect on the refactoring part.
https://github.com/mapswipe/python-mapswipe-workers/blob/5f8c32f12462f43e860536b4ae6a1cfbf684df1e/django/apps/aggregated/management/commands/update_aggregated_data.py#L22-L74


> For backtrack calculation of existing projects your approach is good. I'm just concerned about new projects.

The aggregated module will calculate for old and new projects. Basically, if there are any projects with swipe data then the aggregated module will take care of it.

thenav56: Okay. I have merged this PR and created a new ticket (assigned @ElJocho there)

- https://github.com/mapswipe/python-mapswipe-workers/issues/852

Issue #785, opened 2023-05-17
Integrate new Project Types


ElJocho:
- [x] Media
- [x] Digitization

see #665

PR #806, opened 2023-05-24
Add media project type

ElJocho:
#785 Added Media Type


ElJocho:
-> was merged into project_type_refactoring

PR #807, opened 2023-05-31
Tutorial Workflow

shreeyash07:


frozenhelium:
Already covered in https://github.com/mapswipe/python-mapswipe-workers/pull/860

PR #897, opened 2023-07-18
Split icon array

shreeyash07:


manager-dashboard/app/utils/common.tsx
```diff
@@ -263,11 +257,9 @@ export const iconList: IconItem[] = [
...
         component: IoWarningOutline,
     },
-    {
-        key: 'general-tap',
-        label: 'General Tap',
-        component: getPngIcon(tapIcon, 'general tap'),
-    },
+];
+
+const instructionsIconList: IconItem[] = [
```
tnagorra, line 262: Let's add a NOTE why we have different instructions list

manager-dashboard/app/utils/common.tsx
```diff
@@ -293,10 +285,17 @@ export const iconList: IconItem[] = [
         label: 'Swipe Left',
         component: getPngIcon(swipeIcon, 'swipe left'),
     },
+    {
+        key: 'check',
+        label: 'Check',
+        component: IoCheckmarkOutline,
+    },
```
tnagorra, line 292: Let's rename and move this to iconList.

And also add check from Mapswipe app.

manager-dashboard/app/utils/common.tsx
tnagorra, line 1: Let's rename iconList to customOptionsOnlyIconList

tnagorra (changes requested):

PR #924, opened 2024-01-17
Backend Workers Refactoring

ElJocho:
Big refactoring regarding the project type classes.

Mostly changes from #920.

Additionally it introduces some project types, which are currently not supported in the mapswipe app, but will be relevant in the future and on the upcoming web-client. 

If you have any questions feel free to reach out, so we can discuss the details.


Hagellach37 (approved):
went through these changes

PR #930, opened 2024-05-06
Fix export for group contributor over 500 users

puranban:
### Addresses

- https://github.com/mapswipe/mapswipe/issues/839

### Changes

- Fix user export of group over 500.


community-dashboard/app/components/ItemSelectInput/index.tsx
```diff
@@ -135,17 +141,19 @@ function ItemSelectInput<Name extends string>(props: ItemSelectInputProps<Name>)
 
     const [opened, setOpened] = useState(false);
     const [searchText, setSearchText] = useState<string>('');
-
     const debouncedSearchText = useDebouncedValue(searchText);
 
     const variables = useMemo(() => ({
         search: debouncedSearchText,
+        offset: 0,
+        limit: 5,
```
samshara, line 149: Let's use the defined variable

community-dashboard/app/views/UserGroupDashboard/index.tsx
```diff
@@ -100,6 +104,30 @@ const FILTERED_USER_GROUP_STATS = gql`
...
+                items {
+                    userId
+                    username
+                    isActive
+                    totalMappingProjects
+                    totalSwipeTime
+                    totalSwipes
+                }
+            }
+        }
```
samshara, line 127: we need to add `id` for `userGroup`

puranban: we don't use `id` here, `userId` is used as a unique identifier

frozenhelium: In that case, we probably need to update apollo cache to ise `userID` as key @tnagorra?

puranban: We planned to implement this later.
Reference issue: https://github.com/mapswipe/mapswipe/issues/845

frozenhelium (approved):

PR #941, opened 2024-06-20
Add optional manualUrl field to Manager Dashboard NewProject view

ofr1tz:
The MapSwipe web application can optionally have a button on its project instructions dialog that links to an additional information resource. This PR adds a text input field to the NewProject view of the manager dashboard that allows the project creator to specify a URL that will be written to the `manualUrl` key of the processed project.

The input field is optional, and the user hint clarifies that the link to the additional resource will only be used by the web app.


maze2point0 (approved):
well done

PR #962, opened 2024-08-15
Hotfix/project name validation

puranban:
### Addresses

- https://github.com/mapswipe/mapswipe/issues/872

### Changes

- Validation on project name instead of project title.


manager-dashboard/app/views/NewProject/index.tsx
```diff
@@ -331,8 +331,7 @@ function NewProject(props: Props) {
                 // NOTE: All the user don't have permission to access draft project
                 // FIXME: The firebase rules need to be changed to perform this on draft project
                 const database = getDatabase();
-                const projectTopicKeyLowercase = (projectTopic?.trim())?.toLowerCase() as string;
-                const projectTopicKey = formatProjectTopic(projectTopicKeyLowercase);
+                const projectTopicKey = formatProjectTopic(name);
```
tnagorra, line 334: We should still trim and update case to lowercase.
Is that already done in formatProjectTopic function?

firebase/functions/src/index.ts
```diff
@@ -355,12 +355,14 @@ exports.addProjectTopicKey = functions.https.onRequest(async (_, res) => {
...
-        if (!isEmptyProject && data) {
+        if (data) {
             const newProjectData: {[key: string]: string} = {};
 
             Object.keys(data).forEach((id) => {
-                if (data[id]?.projectTopic) {
-                    const newProjectTopicKey = formatProjectTopic(data[id].projectTopic);
+                const projectData = data[id];
+
+                if (projectData.name) {
```
tnagorra, line 364: We were previously checking if data[id] is undefined. Now, the check has been removed.

firebase/functions/src/index.ts
```diff
@@ -355,12 +355,14 @@ exports.addProjectTopicKey = functions.https.onRequest(async (_, res) => {
             res.status(404).send('No projects found');
         }
 
-        if (!isEmptyProject && data) {
+        if (data) {
```
tnagorra, line 358: Any reason this check was removed?

manager-dashboard/app/views/NewProject/index.tsx
puranban: yes that is already done in `formatProjectTopic` function.

firebase/functions/src/index.ts
puranban: we don't need to check here because project  empty  condition is already handled above.

firebase/functions/src/index.ts
puranban: we checked data[id].projectTopic not data[id],
Now we little simplified by assigning on the another variable.

firebase/functions/src/index.ts
```diff
@@ -355,12 +355,14 @@ exports.addProjectTopicKey = functions.https.onRequest(async (_, res) => {
             res.status(404).send('No projects found');
```
tnagorra, line 355: We might still need to return from this function.

cc: @frozenhelium

firebase/functions/src/index.ts
tnagorra: The analysis is not correct.
The use of `?` means both `data[id]` and `data[id].projectTopic` was checked.

.github/workflows/actions.yml
```diff
@@ -2,6 +2,9 @@ name: MapSwipe Workers
 
 on: [push]
 
+env:
+  COMPOSE_FILE: docker-compose.yaml:docker-compose-ci.yaml
+
```
tnagorra, line 7: @thenav56 Can you verify this usage?

thenav56: Yes, You can check this for more detail https://github.com/mapswipe/python-mapswipe-workers/pull/962/commits/228a85ba60a409dded84b817cf6cccd1eb3fa1e9

firebase/functions/src/index.ts
frozenhelium: Calling `status` will terminate the function. So, we don't need a return here.

tnagorra: Does it throw?

puranban: done

tnagorra (approved):

samshara (approved):

Issue #687, opened 2023-01-27
American Red Cross user group has a user in the member list with no profile name


udaynwa:

Katbaloo:
@kopitek8 any update on this? Still not clear if this is solely an OSM name issue, but given reports from our users, suspect not?  Thanks!

kopitek8:
Hey @Katbaloo. We are working on the fix and should be complete soon. It is an OSM name issue because OSM usernames are not stored in the database while MapSwipe account usernames are. We'll share more if we find any inconsistencies.

kopitek8:
### Solution

We have updated the logic to display the usernames in the User Group table and export. Now if a user's `username` is empty, then the user's ID will be shown. Screenshot of a test OSM user attached below. If more details are required, clicking on the ID will redirect to the user's profile.

### **User Group Members Table**
![Image](https://github.com/user-attachments/assets/522fc119-de36-44d1-9a2c-07abeddfc10f)

### **Profile Page**
![Image](https://github.com/user-attachments/assets/770600d8-f8b4-4636-889a-50789f1f7674)

danbjoseph:
do we have the actual osm username? instead of `osm:*` with the user id number?
for example if i create an account with my osm account i would want it to show `danbjoseph` and not `osm:1535212`

kopitek8:
@danbjoseph 

For the community dashboard, the actual OSM usernames are not stored in the postgres database. Maybe we can add it in future iteration of the python-mapswipe-workers. For the time being, instead of showing empty cells, we decided to at least show usernames.

Whereas, the app pulls data from Firebase where the username is stored.

All in all, we should definitely streamline the stats and data across platforms to make it consistent.

PR #973, opened 2024-10-23
Fix username display on user group dashboard table

puranban:
### Addresses:

- https://github.com/mapswipe/python-mapswipe-workers/issues/687

### Changes

- Fix user name display in the member list for OSM users on user group dashboard


frozenhelium (approved):

tnagorra (changes requested):
There are also other places where username has been used.
Eg: ItemSelectInput, UserGroupDashboard export logic, UserDashboard heading

frozenhelium (approved):

PR #983, opened 2024-12-12
feat: add tutorial for street project type

Gigaszi:


mapswipe_workers/tests/fixtures/tutorialDrafts/street.json
```diff
@@ -0,0 +1,170 @@
+{
+  "createdBy": "LtCUyou6CnSSc1H0Q0nDrN97x892",
+  "tutorialDraftId": "waste_mapping_dar_es_salaam",
+  "taskImageIds": [
```
ofr1tz, line 4: For consistency with tutorial drafts of other project types, it would good to have taskImageIds and answers under `tutorialTasks` in the following structure:

```
"tutorialTasks": {
// crs will be ignored
    "crs": {
      "properties": {   },
      "type": "name"
    },
    "features": [
      {
      // geometry will be ignored
        "geometry": {
          "coordinates": [ ... ],
          "type": "Point"
        },
        "properties": {
          "id": 1543, // = Mapillary image id
          "reference": 1, // correct answer option value
          "screen": 1 // order 
        },
        "type": "Feature"
      },
      { // more features with props... },
      ],
}
```

ofr1tz (approved):
Works for me. Correctly creates a tutorial from a draft made with #1006

PR #988, opened 2025-01-16
fix(manager-dashboard): correctly set isPano for street projects

ofr1tz:
The Mapillary image filter section on the new project form for projects of type Street has a checkbox input to determine whether Mapillary imagery should be filtered for 360° panoramic images or not.

Before this fix, leaving the box unchecked would result in adding a `isPano` field with the value `false`  to the project draft, which in return would mean that the python creation worker would filter available imagery for non-panoramic images.

With this fix, `isPano` is **not** added to the project draft if the checkbox is left unmarked, meaning that both panoramic and non-panoramic images are used, as was originally intended.

### DIFF
--- a/manager-dashboard/app/views/NewProject/index.tsx
+++ b/manager-dashboard/app/views/NewProject/index.tsx
@@ -108,7 +108,7 @@ const defaultProjectFormValue: PartialProjectFormType = {
     // maxTasksPerUser: -1,
     inputType: PROJECT_INPUT_TYPE_UPLOAD,
     filter: FILTER_BUILDINGS,
-    isPano: false,
+    panoOnly: false,
 };
 
 interface Props {
@@ -320,6 +320,7 @@ function NewProject(props: Props) {
 
             valuesToCopy.startTimestamp = valuesToCopy.dateRange?.startDate ?? null;
             valuesToCopy.endTimestamp = valuesToCopy.dateRange?.endDate ?? null;
+            valuesToCopy.isPano = valuesToCopy.panoOnly ? true : null;
 
             const storage = getStorage();
             const timestamp = (new Date()).getTime();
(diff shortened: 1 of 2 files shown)


manager-dashboard/app/views/NewProject/utils.ts
```diff
@@ -83,7 +83,8 @@ export interface ProjectFormType {
     endTimestamp?: string | null;
     organizationId?: number;
     creatorId?: number;
-    isPano?: boolean;
+    panoOnly?: boolean;
+    isPano?: boolean | null;
```
Gigaszi, line 87: shouldnt isPano be removed here?

manager-dashboard/app/views/NewProject/index.tsx
```diff
@@ -320,6 +320,7 @@ function NewProject(props: Props) {
 
             valuesToCopy.startTimestamp = valuesToCopy.dateRange?.startDate ?? null;
             valuesToCopy.endTimestamp = valuesToCopy.dateRange?.endDate ?? null;
+            valuesToCopy.isPano = valuesToCopy.panoOnly ? true : null;
```
Gigaszi, line 323: so it can not be false. or? Is this intended?

ofr1tz: This is as originally intended. The input allows a project creator to decide whether they want all images (isPano = null) or only panoramic ones (isPano = true). In case there should be a need for creating projects with only non-panoramic features, we could adapt the Manager Dashboard later accordingly.

manager-dashboard/app/views/NewProject/utils.ts
ofr1tz: I think this needs to remain as a type check for [valuesToCopy.isPano](https://github.com/mapswipe/python-mapswipe-workers/blob/4811d9426e59fa853798b0136d6904b441200483/manager-dashboard/app/views/NewProject/index.tsx#L323).

Gigaszi (approved):

PR #1006, opened 2025-02-19
Feature/manager dashboard/street tutorial

ofr1tz:
Enables creation of tutorial drafts for projects of type Street from the Manager Dashboard.

The PR includes a sample Street tutorial screens/scenario GeoJSON, as well as a README with instructions on how such scenario files can be created.

The Scenario Pages section of the New Tutorial view does not include scenario page preview for Street project screens, as the project type is currently web only. Previews for scenario screens of other project types are based on the design of the tutorial in the mobile app version.


Gigaszi (approved):
looks good

PR #1013, opened 2025-03-10
Add time calculation diagram and render images

kopitek8:
- Added time calculation diagram
- Rendered diagram in the diagrams markdown file
- Updated diagrams markdown file to render other images as well


tnagorra:
We need to add a worker that periodically gets data from postgres and updates the data in postgres instead of adding a loop in the postgres db.

tnagorra (approved):

PR #1022, opened 2025-04-01
fix: set factor for time_spent_max_allowed in street projects

### LINKED ISSUE #1017: Set reasonable factor for maximum time spent allowed on Street tasks for Community Dashboard stats
To my understanding, a maximum allowed time spent is applied to each mapping session when calculating the contribution times for the Community Dashboard. This is calculated as number of tasks per mapping session multiplied by a factor that is project type specific (see [here](https://github.com/mapswipe/python-mapswipe-workers/blob/5427e0e76634168a9d2ab468d8a2c0fd9a7a00cb/django/apps/aggregated/management/commands/update_aggregated_data.py#L44C9-L64C54)).

Currently, this factor is not defined for projects of type Street and thus defaults to 1. Therefore, we count a maximum of 1 second of mapping time per task for contributions to Street projects. This seems way to little for projects of this type that often demand for more thorough inspection of the imagery.

ofr1tz:
This sets a value of maximum 65 seconds per task to be counted into user mapping time stats for each completed task group in Street projects. The value is derived from the 95th percentile of mapping time per task in existing Street project contributions. That value is expected to decrease as soon as more contributions to Street projects are received.

Closes #1017

### DIFF
--- a/django/apps/aggregated/management/commands/update_aggregated_data.py
+++ b/django/apps/aggregated/management/commands/update_aggregated_data.py
@@ -7,6 +7,7 @@ from apps.aggregated.models import (
     AggregatedUserStatData,
 )
 from apps.existing_database.models import MappingSession, Project
+
 from django.core.management.base import BaseCommand
 from django.db import connection, models, transaction
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


thenav56 (approved):
LGTM

frozenhelium:
@ofr1tz should we deploy and test this in the dev environment? CC: @thenav56

ofr1tz:
Sure @frozenhelium . I think in order to merge, we would need an approval from someone else than me or @thenav56 .

frozenhelium (approved):

